PR #1, opened 2016-09-01
CORE-8041: added default configuration settings

dennisroberts71:


src/apps/util/config.clj
```diff
@@ -240,55 +241,55 @@
...
   [props config-valid configs]
-  "apps.email.base-url")
+  "apps.email.base-url" "http://iplant-email:60000")
 
-(cc/defprop-str app-deletion-notification-src-addr
+(cc/defprop-optstr app-deletion-notification-src-addr
   "The source email address of app deletion notification messages."
   [props config-valid configs]
-  "apps.email.app-deletion.from")
+  "apps.email.app-deletion.from" "prod-user@cyverse.org")
```
psarando: Not sure about having this default email coming from our domain. This means everyone who runs their own DE will be sending app deletion messages to their users from this address.
The current `ansible/inventories/group_vars/all` just defaults this address to a blank string, but I'm not sure if that's appropriate either.
Maybe this shouldn't have any default? Or something more obvious, like `de-notifications@example.com` or `no-reply-please-update-your-de-apps-config@cyverse.org`? 😜

dennisroberts71: Hehe, this is the annoying part of setting configuration defaults. I chose to have it use our domain for a default because it would be more convenient for us, with the idea that other users will have to override this setting. I'll make it a required setting instead.

psarando:
Looks good to me :shipit: 👍

dennisroberts71:
Thanks for the review.

PR #2, opened 2016-09-02
CORE-5372 Remove “no tool” error in /apps/:id/details endpoint.

psarando:
The UI is able to display app details with an empty "Tool Information" tab for private apps that have no tools set yet.

So this error response is obsolete.


ianmcorvidae:
The change looks fine, and I tried to briefly check that this wouldn't just result in schema validation issues, but it looks probably fine to (since this'll result in an empty list of tools, not a missing `:tools` key). So I'd say :shipit:

psarando:
Thanks for the review 👍 
I did test this change locally and the response validates, and the UI is able to display empty tool info for an app without tools.

PR #3, opened 2016-09-08
Add the execution system ID to analysis and app listing endpoints.

dennisroberts71:
This adds the job execution system ID, which is different from Agave's execution system iD, to several app and analysis editing and listing endpoints.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -24,7 +24,7 @@
                  [org.cyverse/authy "2.8.0"]
                  [org.cyverse/clojure-commons "2.8.0"]
                  [org.cyverse/kameleon "2.8.0"]
-                 [org.cyverse/mescal "2.8.0"]
+                 [org.cyverse/mescal "2.8.1-SNAPSHOT"]
                  [org.cyverse/metadata-client "2.8.0"]
                  [org.cyverse/common-cli "2.8.0"]
                  [org.cyverse/common-cfg "2.8.0"]
--- a/src/apps/service/apps/agave/jobs.clj
+++ b/src/apps/service/apps/agave/jobs.clj
@@ -119,6 +119,7 @@
     :batch           false
     :description     (:description job)
     :enddate         (:enddate job)
+    :system_id       "agave"
     :id              job-id
     :name            (:name job)
     :notify          (:notify job)
--- a/src/apps/service/apps/agave/jobs.clj
+++ b/src/apps/service/apps/agave/jobs.clj
@@ -119,7 +119,7 @@
     :batch           false
     :description     (:description job)
     :enddate         (:enddate job)
-    :system_id       "agave"
+    :system_id       jp/agave-client-name
     :id              job-id
     :name            (:name job)
     :notify          (:notify job)
--- a/src/apps/service/apps/job_listings.clj
+++ b/src/apps/service/apps/job_listings.clj
@@ -50,7 +50,7 @@
     :app_name        (:app-name job)
     :description     (:description job)
     :enddate         (job-timestamp (:end-date job))
-    :system_id       (string/lower-case (:job_type job))
+    :system_id       (string/lower-case (:job-type job))
     :id              id
     :name            (:job-name job)
     :resultfolderid  (:result-folder-path job)
--- a/src/apps/service/apps/job_listings.clj
+++ b/src/apps/service/apps/job_listings.clj
@@ -42,6 +42,8 @@
   [apps-client rep-steps {:keys [parent-id id]}]
   (and (nil? parent-id) (job-permissions/job-steps-support-job-sharing? apps-client (rep-steps id))))
 
+(def job-type-to-system-id string/lower-case)
+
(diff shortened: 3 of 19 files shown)


src/apps/service/apps/agave/jobs.clj
```diff
@@ -119,6 +119,7 @@
     :batch           false
     :description     (:description job)
     :enddate         (:enddate job)
+    :system_id       "agave"
```
ianmcorvidae: Pretty minor and I only thought of it because I just reviewed the mescal PR, but might it make sense to pull this from `mescal.agave-de-v2.constants/hpc-system-id`? Seems like an easily-overlooked lone place it might need changing, if for some reason it needs changing.

src/apps/service/apps/job_listings.clj
```diff
@@ -45,6 +50,7 @@
     :app_name        (:app-name job)
     :description     (:description job)
     :enddate         (job-timestamp (:end-date job))
+    :system_id       (string/lower-case (:job-type job))
```
ianmcorvidae: Will the system ID always be the lowercased job type? If so it seems like a lot of code could be condensed/made dependent on this fact. Minor, but might be worth considering if we're enforcing this.

project.clj
```diff
@@ -24,7 +24,7 @@
                  [org.cyverse/authy "2.8.0"]
                  [org.cyverse/clojure-commons "2.8.0"]
                  [org.cyverse/kameleon "2.8.0"]
-                 [org.cyverse/mescal "2.8.0"]
+                 [org.cyverse/mescal "2.8.1-SNAPSHOT"]
```
psarando: Should this `SNAPSHOT` dependency be committed?
Or should this just be `2.8.1` after this version of the mescal lib has been released?

src/apps/service/apps/agave/jobs.clj
dennisroberts71: thanks for catching this. I meant to use a separate constant here rather than the value from `mescal.agave-de-v2.constants`

project.clj
dennisroberts71: My plan was to use `2.8.1-SNAPSHOT` for the time being then change it when we get ready to release the code.

src/apps/service/apps/job_listings.clj
dennisroberts71: It is currently, but I wasn't planning to rely on this fact.

project.clj
psarando: Ah, yeah, sorry, I didn't notice that mescal was already at `2.8.1-SNAPSHOT`.
I was thinking it was still at `2.8.0`.

dennisroberts71: Cool, no worries.

ianmcorvidae: I agree with @dennisroberts71 on this one -- obviously we want to avoid snapshot dependencies in releases of services, but I think it's fine to commit them in snapshot versions while testing is ongoing. Since there may easily be more changes to mescal before this work is done, it makes sense to not release it until it's a reasonably complete package of changes (unless a release is needed for some other reason before then, of course).

ianmcorvidae:
My comments are mostly food for thought; the basic idea and etc. looks good, so I'd say :shipit: as you're ready!

src/apps/service/apps/job_listings.clj
```diff
@@ -58,7 +64,7 @@
     :parent_id       parent-id
     :batch           (:is-batch job)
     :batch_status    (when (:is-batch job) (format-batch-status id))
-    :can_share       (and (nil? parent-id) (job-permissions/job-steps-support-job-sharing? apps-client (rep-steps id)))}))
+    :can_share       (job-supports-sharing? apps-client rep-steps job)}))
```
psarando: This doesn't seem related to system IDs.
Is this a bug-fix? Or just a minor refactor?

dennisroberts71: Just a minor refactor. My long line OCD got the best of me. 😛

psarando:
I didn't catch any issues with these changes.
:shipit:

dennisroberts71:
Cool, thanks for the reviews!

src/apps/service/apps/job_listings.clj
ianmcorvidae: Perhaps for future-proofing, it'd make sense to factor it out as a variable which can be used as a function, mapping job type to system ID. For now, it can just be something like `(def job-type-to-system-id string/lower-case)`, but if later on it gets more complex it can be changed to a map or a more complex function, and that way it only needs changing one place.

project.clj
ianmcorvidae: (heh, left my comment before the two comments above it. c'est la vie)

src/apps/service/apps/job_listings.clj
dennisroberts71: Yeah, that would be useful.

PR #5, opened 2016-09-17
CORE-8065: add the system ID to apps URL paths.

dennisroberts71:
This is a work in progress. Three endpoints using qualified app IDs have been implemented: `POST /apps/:system-id`, `POST /apps/:system-id/arg-preview`, and `GET /apps/:system-id/:app-id`. Code not called by these endpoints will be subject to further modification as new versions of their endpoints are created and doesn't need to be reviewed as thoroughly at this time.


dennisroberts71:
I'm closing this PR so that I can open a new one against the master branch later this week.

PR #6, opened 2016-09-20
Add the execution system ID to apps endpoints.

dennisroberts71:
The current way of routing calls to different execution systems (e.g. Agave, vs the DE's HTCondor cluster) is brittle because it relies on the fact that DE app IDs are UUIDs whereas Agave's are not. This change adds the system ID to the URI path so that calls can be routed in a much more extensible manner.

This change is a work in progress. Code that has ben aded for endpoints that haven't been converted yet is subject to change further changes.


src/apps/service/apps/agave.clj
```diff
@@ -68,33 +74,60 @@
...
+
   (formatPipelineTasks [_ pipeline]
     (pipelines/format-pipeline-tasks agave pipeline))
 
   (listJobs [self params]
     (job-listings/list-jobs self user params))
 
+  ;; TODO: Determine how this should be refactored now that we have system IDs. If I remember correctly,
+  ;; this method is used during job listings. If that's the case, we can filter apps by execution system
+  ;; easily enough.
```
strootman: If there is not a ticket for this, there should be.

src/apps/service/apps/agave.clj
```diff
@@ -105,6 +138,10 @@
     (when-not (util/uuid? (:app_id submission))
       (agave-jobs/submit agave user submission)))
 
+  (submitJob [this system-id submission]
+    (validate-system-id system-id)
+    (agave-jobs/submit agave user submission))
+
```
strootman: Not having to check that the `app_id` is _not_ a UUID is a good change!

src/apps/service/apps/agave.clj
```diff
@@ -139,61 +180,106 @@
...
+    (validate-system-id system-id)
+    (reject-app-documentation-edit-request))
+
   (adminAddAppDocs [_ app-id _]
     (when-not (util/uuid? app-id)
       (reject-app-documentation-edit-request)))
 
+  (adminAddAppDocs [_ system-id app-id _]
+    (validate-system-id system-id)
+    (reject-app-documentation-edit-request))
```
strootman: Are there service tests for these endpoints which return `bad requests`? Seems like easy pickings.

strootman:
This PR looks LGTM. The corresponding ticket may be difficult to validate in its current state. I think it is necessary to list every endpoint affected/added. Also, the endpoints will need to be added to the drop notes as well.

src/apps/service/apps/agave.clj
dennisroberts71: I was going to do this under the same ticket. The comment is just a reminder for myself.

src/apps/service/apps/agave.clj
dennisroberts71: There aren't yet, but there will be once the admin endpoints are converted.

dennisroberts71:
Thanks for the review! I'll create subtasks for each of the individual repositories for extra Jirosis.

strootman:
We discussed this offline, but adding it here.
I didn't mean to add one jira per endpoint, just that the endpoints be listed in CORE-8065 (or where appropriate). This was only to facilitate the ability to properly validate the ticket/work.

strootman:
Last thing. Be sure to rebase your changes to the master branch and force push your branch before merging this PR.

PR #7, opened 2016-10-12
CORE-8162 Add job stats fields to app listings and details endpoints.

psarando:
This PR adds a `job_stats` JSON object with `job_count_completed` and `job_last_completed` fields to the following endpoint responses:
* `GET /apps`
* `GET /apps/{app-id}/details`
* `GET /apps/{system-id}/{app-id}/details`
* `GET /apps/categories/{category-id}`
* `GET /apps/hierarchies/{root-iri}/apps`
* `GET /apps/hierarchies/{root-iri}/unclassified`

For example:
```json
{
  "job_stats": {
    "job_count_completed": 0,
    "job_last_completed": "2016-10-12T00:38:57.037Z"
  }
}
```

Also a `job_stats` field with the same JSON object above, but also with `job_count`, `job_count_failed`, and `last_used` fields were added to the following endpoint responses:
* `PATCH /admin/apps/{app-id}`
* `GET /admin/apps/{app-id}/details`
* `GET /admin/ontologies/{ontology-version}/{root-iri}/apps`
* `GET /admin/ontologies/{ontology-version}/{root-iri}/unclassified`

For example:
```json
{
  "job_stats": {
    "job_count_completed": 0,
    "job_last_completed": "2016-10-12T00:38:57.037Z",
    "job_count": 0,
    "job_count_failed": 0,
    "last_used": "2016-10-12T00:38:57.037Z"
  }
}
```

The updated listing endpoints are also able to sort results on the fields added to their responses.

### DIFF
--- a/src/apps/service/apps/agave/listings.clj
+++ b/src/apps/service/apps/agave/listings.clj
@@ -9,15 +9,9 @@
 
 (defn- add-agave-job-stats
   [{:keys [id] :as app} admin?]
-  (let [job-stats     (if admin?
-                        (jobs-db/get-job-stats id)
-                        (jobs-db/get-public-job-stats id))
-        default-stats (if admin?
-                        {:job_count 0
-                         :job_count_failed 0
-                         :job_count_completed 0}
-                        {:job_count_completed 0})]
-    (merge app default-stats job-stats)))
+  (merge app (if admin?
+               (jobs-db/get-job-stats id)
+               (jobs-db/get-public-job-stats id))))
 
 (defn- add-app-listing-job-stats
   [app-listing admin?]
--- a/src/apps/service/apps/agave/listings.clj
+++ b/src/apps/service/apps/agave/listings.clj
@@ -1,24 +1,57 @@
 (ns apps.service.apps.agave.listings
-  (:use [apps.service.util :only [sort-apps apply-offset apply-limit uuid?]]
+  (:use [apps.service.util :only [sort-apps apply-offset apply-limit format-job-stats uuid?]]
+        [apps.util.conversions :only [remove-nil-vals]]
         [slingshot.slingshot :only [try+]])
   (:require [clojure.tools.logging :as log]
             [clojure-commons.error-codes :as ce :refer [clj-http-error?]]
-            [apps.persistence.app-metadata :as ap]))
+            [apps.persistence.app-metadata :as ap]
+            [apps.persistence.jobs :as jobs-db]))
+
+(defn- add-agave-job-stats
+  [{:keys [id] :as app} admin?]
+  (let [job-stats     (if admin?
+                        (jobs-db/get-job-stats id)
+                        (jobs-db/get-public-job-stats id))
+        default-stats (if admin?
+                        {:job_count 0
+                         :job_count_failed 0
+                         :job_count_completed 0}
+                        {:job_count_completed 0})]
+    (merge app default-stats job-stats)))
+
+(defn- add-app-listing-job-stats
+  [app-listing admin?]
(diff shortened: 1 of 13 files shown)


src/apps/service/apps/agave/listings.clj
```diff
@@ -1,24 +1,58 @@
...
+  (log/spy id)
+  (let [job-stats     (if admin?
+                        (jobs-db/get-job-stats id)
+                        (jobs-db/get-public-job-stats id))
+        default-stats (if admin?
+                        {:job_count 0
+                         :job_count_failed 0
+                         :job_count_completed 0}
+                        {:job_count_completed 0})]
+    (merge app default-stats job-stats)))
```
psarando: If an Agave app has never been run by a DE user, then that app will not have any stats returned by these queries, which will cause that app to fail response validation. So these required, default fields are set in case these queries return `nil`.

psarando: This is not an issue for DE app listings, since the `app_listing` view always includes at least a `0` value for these fields for DE apps that have not run yet.

src/apps/service/apps/agave/listings.clj
```diff
@@ -1,24 +1,58 @@
...
+  [agave app-id admin?]
+  (-> (.getAppDetails agave app-id)
+      (add-agave-job-stats admin?)
+      (format-job-stats admin?)
+      remove-nil-vals))
 
 (defn list-apps
   [agave category-id params]
   (-> (.listApps agave)
+      (add-app-listing-job-stats false)
```
psarando: Unfortunately we have to query the job stats of every Agave app in the initial listing, before applying limits and offsets, in case the request wishes to sort the results by one of the job stats fields.

src/apps/service/util.clj
```diff
@@ -11,8 +11,8 @@
   (partial sort-by
            (keyword sort-field)
            (if (and sort-dir (= (string/upper-case sort-dir) "DESC"))
-             #(compare (string/lower-case %2) (string/lower-case %1))
-             #(compare (string/lower-case %1) (string/lower-case %2)))))
+             #(compare (string/lower-case (str %2)) (string/lower-case (str %1)))
+             #(compare (string/lower-case (str %1)) (string/lower-case (str %2))))))
```
psarando: This `app-sorter` function was throwing an NPE for fields with non-string values. This bug probably wasn't caught before now since the UI currently does not page app listings, and so it currently does all its own sorting in the browser.

psarando:
Yep, I traced everywhere the app_listing view is used.
Mostly is used when fetching info for a single app, but in those cases most fields (including the new job stats fields) are not used and filtered out in any endpoint responses.

These new fields are only returned in the apps details endpoints and listing/search endpoints.

src/apps/service/apps/agave/listings.clj
```diff
@@ -27,12 +54,14 @@
...
-  [agave search-term params]
+  [agave search-term params admin?]
   (try+
    (-> (.searchApps agave search-term)
+       (add-app-listing-job-stats admin?)
        (sort-apps params {:default-sort-field "name"})
        (apply-offset params)
-       (apply-limit params))
+       (apply-limit params)
+       (format-app-listing-job-stats false))
```
dennisroberts71: Should this `false` be `admin?` instead?

psarando: You're right, this should be `admin?`.
That also reminded me that I haven't added an `/admin/apps/` endpoint to use these admin search functions yet 😊

psarando:
Thanks for the reviews!

PR #10, opened 2016-10-20
CORE-8162 Remove job_stats from non-admin app listing endpoint results

psarando:
Removes `job_stats` from non-admin app listing endpoint results, since job stat columns have been removed from the `app_listing` view by cyverse-de/de-db#2.

Admin app search and listing endpoints now use the new admin search and app listing functions added by cyverse-de/kameleon#2 so they can still include `job_stats` in their results.

Also includes a fix for adding admin `job_stats` to `apps.service.apps.agave.listings/list-apps-with-ontology` results.

### DIFF
--- a/src/apps/routes/schemas/app.clj
+++ b/src/apps/routes/schemas/app.clj
@@ -396,9 +396,6 @@
      :step_count
      (describe Long "The number of Tasks this App executes")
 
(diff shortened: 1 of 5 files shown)


psarando:
Thanks for the review!

PR #11, opened 2016-10-21
CORE-8162 Exclude parent jobs from app details `job_stats` counts.

psarando:
This PR will exclude the parent job of batch jobs in the `job_stats` counts of app details endpoint responses.


psarando:
Thanks. The apps service is already using kameleon "2.8.2-SNAPSHOT", which is the current version. I might do a release version of kameleon now and update the version in this PR before I merge it, though.

PR #12, opened 2016-10-21
CORE-3949 Exclude non-public apps from admin app search results.

psarando:
This PR allows admins to restrict app searches to only public apps.

### DIFF
--- a/src/apps/service/apps/de/listings.clj
+++ b/src/apps/service/apps/de/listings.clj
@@ -40,17 +40,20 @@
    (augment-listing-params params short-username (perms-client/load-app-permissions short-username))))
 
 (defn- augment-search-params
-  [search_term params short-username app-ids]
-  (let [category-attrs (set (workspace-metadata-category-attrs))]
-    (if (or (empty? app-ids) (empty? category-attrs))
-      params
-      (assoc params :pre-matched-app-ids
-                    (metadata-client/filter-targets-by-ontology-search short-username
-                                                                       (categorization/get-active-hierarchy-version)
-                                                                       category-attrs
-                                                                       search_term
-                                                                       ["app"]
-                                                                       app-ids)))))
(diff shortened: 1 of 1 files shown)


src/apps/service/apps/de/listings.clj
```diff
@@ -40,17 +40,20 @@
...
-                                                                       (categorization/get-active-hierarchy-version)
-                                                                       category-attrs
-                                                                       search_term
-                                                                       ["app"]
-                                                                       app-ids)))))
+  [search_term {:keys [app-ids public-app-ids] :as params} short-username admin?]
+  (let [category-attrs (set (workspace-metadata-category-attrs))
+        app-ids        (if admin? public-app-ids app-ids)]
+    (-> params
+        (assoc :app-ids             app-ids
```
psarando: The app search query in `kameleon` only searches apps with IDs in this `app-ids` list.
By setting the `app-ids` param to the list of `public-app-ids`, admin searches are restricted to only public apps.

src/apps/service/apps/de/listings.clj
```diff
@@ -40,17 +40,20 @@
...
+               :pre-matched-app-ids (when-not (or (empty? app-ids) (empty? category-attrs))
```
dennisroberts71: This is an extremely trivial suggestion, but I'd probably write something like:

```
(when (every? seq [app-ids category-attrs]) ...)
```

or

```
(when-not (some empty? [app-ids category-attrs]) ...)
```

It's not functionally different, though; it's just a personal preference. Feel free to ignore this suggestion, of course. :)

psarando: I forget to use `some` and `every?` more often.
Since it's only 2 items, I think I'll leave it this way for now.
Thanks for the tip!

PR #14, opened 2016-10-28
CORE-8375 Fixes for apps listing endpoints.

psarando:
This PR introduces 2 major changes for app listing responses:

1.  Allows sorting results under the `job_stats` and `rating` fields and fixes combined app sorting for fields with number values (such as job counts).
2. Renames the `app_count` field to `total` since this is what the UI expects for paging results.

The swagger docs of the `GET /admin/apps` and `GET /apps` endpoints were also updated to enumerate their valid sort fields.

This PR also requires changes from cyverse-de/mescal#2.

### DIFF
--- a/src/apps/routes/schemas/app.clj
+++ b/src/apps/routes/schemas/app.clj
@@ -1,6 +1,6 @@
 (ns apps.routes.schemas.app
   (:use [common-swagger-api.schema :only [->optional-param
-                                          ->required-key
+                                          optional-key->keyword
                                           describe
                                           PagingParams
                                           SortFieldDocs
@@ -422,7 +422,7 @@
          {:apps (describe [AdminAppListingDetail] "A listing of App details")}))
 
 (def AppListingValidSortFields
-  (-> (map ->required-key (keys AppListingDetail))
+  (-> (map optional-key->keyword (keys AppListingDetail))
       (conj :average_rating :user_rating)
       set
       (sets/difference #{:app_type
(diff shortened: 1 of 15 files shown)


src/apps/routes/schemas/app.clj
```diff
@@ -415,6 +421,50 @@
   (merge AppListing
          {:apps (describe [AdminAppListingDetail] "A listing of App details")}))
 
+(def AppListingValidSortFields
+  (-> (map ->required-key (keys AppListingDetail))
```
dennisroberts71: It might be helpful to create an alias of `->required-key` indicating that the schema key is being converted to a keyword. Something like `schema-key->keyword`.

src/apps/routes/schemas/app.clj
```diff
@@ -415,6 +421,50 @@
...
+(defschema AppListingPagingParams
+  (merge SecuredQueryParamsEmailRequired
+         PagingParams
+         {SortFieldOptionalKey
+          (describe (apply enum AppListingValidSortFields) SortFieldDocs)}))
+
+(def AppSearchValidSortFields
+  (-> AppListingValidSortFields
+      (sets/difference #{:average_rating :user_rating})
+      (conj :average :total)))
```
dennisroberts71: Is `:total` used to sort by the total rating?

src/apps/service/util.clj
```diff
@@ -1,18 +1,38 @@
...
             [clojure-commons.exception-util :as cxu]
             [kameleon.uuids :as uuids])
   (:import [java.util UUID]))
 
+(defn- app-sorter-keyfn
+  [sort-field]
+  (let [sort-field (keyword sort-field)]
+    (condp contains? sort-field
+      AdminAppListingJobStatsKeys (comp sort-field :job_stats)
+      #{:average :total}          (comp sort-field :rating)
```
dennisroberts71: Ah, I answered my own question from above. I don't know if it's possible, but if enumeration values can have descriptions, it might be helpful to add some short descriptions.

psarando: I don't think it's possible to display descriptions for these enums in the swagger docs, since these values are only displayed in the drop-down list for the query param.

psarando:
@dennisroberts71 Thanks for the review! Would you mind taking a quick look at cyverse-de/mescal#2 also?

dennisroberts71:
No problem. I'll take a look at the other PR now.

psarando:
Thanks again for the review!

PR #15, opened 2016-10-28
CORE-8401: ensure that the location is included in all redirects retu…

dennisroberts71:
…rned by the apps service

### DIFF
--- a/src/apps/service/oauth.clj
+++ b/src/apps/service/oauth.clj
@@ -69,15 +69,25 @@
 (defn authorization-uri
   "Generates an authorization URI for a remote API."
   [{:keys [api-name] :as server-info} username state-info]
-  (when-not (op/get-access-token api-name username)
-    (str (assoc (curl/url (:auth-uri server-info))
-                :query {:response_type "code"
-                        :client_id     (:client-key server-info)
-                        :redirect_uri  (:redirect-uri server-info)
-                        :state         (op/store-authorization-request username state-info)}))))
+  (str (assoc (curl/url (:auth-uri server-info))
+              :query {:response_type "code"
+                      :client_id     (:client-key server-info)
+                      :redirect_uri  (:redirect-uri server-info)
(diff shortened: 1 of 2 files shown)


src/apps/service/oauth.clj
```diff
@@ -69,15 +69,25 @@
 (defn authorization-uri
   "Generates an authorization URI for a remote API."
   [{:keys [api-name] :as server-info} username state-info]
-  (when-not (op/get-access-token api-name username)
```
dennisroberts71: The problem was that `authorization-uri` was returning nil when the user already had an access token even if the access token wasn't any good. `authorization-uri` was called (here)[https://github.com/dennisroberts71/apps/blob/master/src/apps/service/apps.clj#L22-L25] to set the location header in the redirect.

src/apps/service/oauth.clj
```diff
@@ -69,15 +69,25 @@
...
+
+(defn has-access-token
+  "Determines whether or not a user has an access token for an external API."
+  [{:keys [api-name] :as server-info} username]
+  (seq (op/get-access-token api-name username)))
+
+(defn- build-auth-uri
+  "Builds an authorization URI for an external API if the user doesn't have an access token already."
+  [server-info username]
+  (when-not (has-access-token server-info username)
```
dennisroberts71: It's not really new, just refactored a little bit. The call to `has-access-token` serves the same purpose as the call to `op/get-access-token` did above. I just moved it out of `authorization-uri`.

psarando: Ah, yep, I see that now (again), thanks 😁

src/apps/service/oauth.clj
psarando: Ah, it must be a Friday afternoon...
So `authorization-uri` no longer does a `when-not` check so it will no longer return nil 💡

dennisroberts71:
Thanks for the reviews!

PR #17, opened 2016-11-09
CORE-7813 Add auth_redirect to /users/login response.

psarando:
This PR combines user login responses for the terrain `bootstrap` endpoint, so it only needs to make 1 call to get both pieces of user "session" info.

This info wasn't combined with the apps `GET /workspaces` endpoint response, since this kind of user info will eventually be moved into a new `user-info` service, and workspaces will likely remain apps service specific.


psarando:
Thanks for the review 👍

PR #18, opened 2016-11-23
CORE-8421 Migrate apps specific queries and entities out of the kameleon library

psarando:
The kameleon library is used by a number of our services for setting up database connections and running database conversions.

Some functions and entities defined in kameleon are specific to the apps service and are not used elsewhere, so they have been migrated out of the kameleon library (see cyverse-de/kameleon#6) and into this apps service.


src/apps/service/apps/de/jobs/common.clj
```diff
@@ -1,17 +1,15 @@
 (ns apps.service.apps.de.jobs.common
-  (:use [kameleon.core]
-        [kameleon.entities]
-        [kameleon.queries :only [get-user-id]]
+  (:use [apps.persistence.users :only [get-user-id]]
+        [apps.util.assertions :only [assert-not-nil]]
+        [apps.util.conversions :only [remove-nil-vals]]
```
dennisroberts71: Heh, another case where I'd like to see `clojure-commons.core/remove-nil-values` used instead.

src/apps/service/apps/de/pipeline_edit.clj
```diff
@@ -1,22 +1,21 @@
...
+                                              add-mapping
+                                              add-step
+                                              add-task
+                                              get-app
+                                              remove-app-steps
+                                              update-app]]
+        [apps.persistence.entities :only [app_steps input_mapping]]
+        [apps.service.apps.de.edit :only [add-app-to-user-dev-category app-copy-name]]
         [apps.service.apps.de.validation :only [verify-app-editable]]
+        [apps.util.conversions :only [remove-nil-vals]]
```
dennisroberts71: Here's another call to `remove-nil-vals`. These updates don't really have to be done now. It was just something that I've been meaning to change for a long time.

psarando: Yep. I think I'll keep it out of this PR (since this PR just rearranged the `:require` deps and didn't introduce this function to this namespace or service).

But it's a good idea for another commit/PR after this PR.

src/apps/persistence/workspace.clj
```diff
@@ -24,16 +26,34 @@
        (map-indexed (partial app-groups/add-subgroup root-category-id))
        (dorun)))
 
+(defn set-workspace-root-app-group
```
psarando: Good catch on this and the dead code.
Did you run a linter to find these? My `lein eastwood` linter didn't pick these up.

dennisroberts71: Thanks. No, the linter is configured not to look for dead code because it was reporting false positives for functions that were only called from within one of the `deftype` definitions. I found that the functions weren't being used by running `git grep` for each function.

psarando:
Thanks for the review 👍

PR #19, opened 2016-12-02
CORE-8455: pass docker registry auth information down to jex-adapter etc.

ianmcorvidae:
Since the model namespace has been updated, this gets passed the whole way to road-runner. Which, right now, doesn't actually use it. That's next, after cyverse-de/dockerops#1 can be used to do it.

Tested that everything still works properly with this (and made a few small changes to other code to ensure as much).

### DIFF
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -9,6 +9,7 @@
                                           container-volumes
                                           container-volumes-from
                                           data-containers]]
+        [apps.persistence.docker-registries :only [get-registry]]
         [apps.util.assertions :only [assert-not-nil]]
         [apps.util.conversions :only [remove-nil-vals remove-empty-vals]]
         [apps.validation :only [validate-image-not-public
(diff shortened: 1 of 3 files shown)


src/apps/containers.clj
```diff
@@ -18,7 +19,8 @@
         [korma.core :exclude [update]]
         [korma.db :only [transaction]])
   (:require [clojure.tools.logging :as log]
-            [korma.core :as sql]))
+            [korma.core :as sql]
+            [clojure.string :as string]))
```
psarando: I've been trying to keep `:require` lists in alphabetical order, especially once they get long.

johnworth:
Looks fine as far as I can tell. :shipit:

ianmcorvidae:
Cool, thanks for taking a look

src/apps/containers.clj
ianmcorvidae: I'll switch it around; I hadn't realized that was the intention.

psarando: It's not consistent everywhere. In apps there are places where I started mimicking the import order used in our UI code, so `clojure.*` comes first, 3rd party libs next, then `apps.*` at the bottom, but I think that's more confusing in clojure than just alphabetical order 😖

PR #20, opened 2016-12-07
CORE-8119: fixed some apps integration tests that had been commented out

dennisroberts71:


ianmcorvidae:
The changes look basically straightforward to me. I'm guessing since this modifies things, we're not running it against live/"real" environments where this sort of thing would have the potential for causing issues (I'm not sure what would happen if this text fixture ran against e.g. de-2). So I'd say :shipit:

dennisroberts71:
Thanks for the review!

psarando:
These changes look simple enough.

This will leave cruft in the db's ontologies table of long running environments (deleting the ontology leaves the file in the db, but sets the `deleted` field to `true` so it's not displayed as an option in the DE), so these tests shouldn't be run against anything like prod where the db isn't reset periodically.

I'm also not sure if we're allowed to redistribute this OWL Ontology file, so we should double check the EDAM license.

dennisroberts71:
These tests are designed to run against a crash-and-burn database, so the cruft shouldn't be a problem. I'm glad you mentioned the licensing issue; that hadn't occurred to me. I checked, and EDAM uses this license: https://creativecommons.org/licenses/by-sa/4.0/. The licensing shouldn't be a problem as far as I can tell.That said, I would eventually like to move away from the current integration testing mechanism that the `apps` service

psarando:
I'm not a license expert, but looking at http://edamontologydocs.readthedocs.io/en/latest/license.html it would seem it's OK for us to include this file in our repo.

PR #21, opened 2016-12-08
CORE-8487 Update to new protocol version of metadata-client

psarando:
Updates for the new protocol version of `metadata-client` added by cyverse-de/metadata-client#1.

Includes a new `apps.clients.metadata` namespace to facade calls to `metadata-client` with common apps service parameters.

### DIFF
--- a/src/apps/clients/metadata.clj
+++ b/src/apps/clients/metadata.clj
@@ -4,23 +4,25 @@
             [apps.util.config :as config]
             [metadata-client.core :as metadata-client]))
 
+(def app-target-type "app")
+
 (defn filter-by-avus
   [username app-ids avus]
-  (metadata-client/filter-by-avus (config/metadata-client) username ["app"] app-ids avus))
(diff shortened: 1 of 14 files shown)


src/apps/clients/metadata.clj
```diff
@@ -0,0 +1,92 @@
+(ns apps.clients.metadata
+  (:use [slingshot.slingshot :only [throw+]])
+  (:require [apps.persistence.categories :as db-categories]
+            [apps.util.config :as config]
+            [metadata-client.core :as metadata-client]))
+
+(defn filter-by-avus
+  [username app-ids avus]
+  (metadata-client/filter-by-avus (config/metadata-client) username ["app"] app-ids avus))
```
dennisroberts71: This is almost too trivial to mention, but it might be helpful to define a variable to hold the target type (just to make sure that a mistyped `"app"` constant doesn't cause problems in the future).

psarando: Good suggestion, I'll add it 👍

psarando:
Thanks for the review 👍

PR #22, opened 2016-12-09
Add the system ID to all non-admin apps endpoints.

dennisroberts71:
The location of the system ID depends on the nature of the endpoint. For many of the endpoints, it appears in the URI path. For endpoints where that doesn't make sense (for example, bulk update endpoints and pipeline endpoints), the system ID will appear in the request body. This is a significant change, and I wouldn't be surprised if there are issues. A thorough review would be appreciated.

### DIFF
--- a/src/apps/routes/apps.clj
+++ b/src/apps/routes/apps.clj
@@ -485,4 +485,14 @@
         the means to store the App rating. This service accepts a rating level between one and
         five, inclusive, and a comment identifier that refers to a comment in iPlant's Confluence
         wiki. The rating is stored in the database and associated with the authenticated user."
-        (ok (apps/rate-app current-user system-id app-id body))))))
+        (ok (apps/rate-app current-user system-id app-id body)))
+
+      (GET "/tasks" []
+        :query [params SecuredQueryParams]
+        :return AppTaskListing
+        :summary "List Tasks with File Parameters in an App"
+        :description "When a pipeline is being created, the UI needs to know what types of files are
+        consumed by and what types of files are produced by each App's task in the pipeline. This
+        service provides that information."
+        (ok (coerce! AppTaskListing
+                     (apps/get-app-task-listing current-user system-id app-id)))))))
--- a/src/apps/routes/apps.clj
+++ b/src/apps/routes/apps.clj
@@ -504,4 +504,13 @@
         :description "This service lists information for all of the tools that are associated with an App.
         This information used to be included in the results of the App listing service."
         (ok (coerce! NewToolListing
-                     (apps/get-app-tool-listing current-user system-id app-id)))))))
+                     (apps/get-app-tool-listing current-user system-id app-id))))
+
+      (GET "/ui" []
+        :query [params SecuredQueryParamsEmailRequired]
+        :return App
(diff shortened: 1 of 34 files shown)


src/apps/service/apps/de/jobs.clj
```diff
@@ -43,32 +39,32 @@
...
                      :status          status
-                     :job-type        "DE"
-                     :app-step-number 1}))
+                     :job_type        jp/de-job-type
+                     :app_step_number 1}))
 
 (defn- save-job-submission
   "Saves a DE job and its job-step in the database."
-  ([user job submission]
+  ([user  job submission]
```
psarando: Trivial: looks a space snuck in here 👾

src/apps/service/apps/de/jobs.clj
```diff
@@ -95,17 +91,26 @@
...
+(defn- submit-job-in-batch
+  [user submission job]
+  (->> (try+
+        (do-jex-submission job)
+        (save-job-submission user job submission)
+        (catch Object _
+          (save-job-submission user job submission "Failed")))
+       (format-job-submission-response user job true)))
+(defn- submit-standalone-job
+
```
psarando: Trivial: Another 👾

src/apps/service/apps/de/jobs.clj
```diff
@@ -95,17 +91,26 @@
...
-     (do-jex-submission job)
-     (catch Object _
-       (if batch?
-         (save-job-submission user job submission "Failed")
-         (throw+))))
-    (->> (save-job-submission user job submission)
-         (format-job-submission-response user job batch?))))
+  (if (:parent_id submission)
+    (submit-job-in-batch user submission job)
+    (submit-standalone-job user submission job)))
```
psarando: Tidy! ✨

src/apps/routes/apps.clj
```diff
@@ -454,4 +454,63 @@
...
+                     (apps/get-app-tool-listing current-user system-id app-id))))
+
+      (GET "/ui" []
+        :query [params SecuredQueryParamsEmailRequired]
+        :return App
+        :summary "Make an App Available for Editing"
+        :description "The app integration utility in the DE uses this service to obtain the App
+        description JSON so that it can be edited. The App must have been integrated by the
+        requesting user."
+        (ok (apps/get-app-ui current-user system-id app-id))))))
```
dennisroberts71: I think they can. I'm planning to remove them all as soon as the rest of the endpoints are updated. There will be a significant amount of dead code to remove at that time, so I'll probably do that after the new updates are all in place so that we have a working, if someone bloated, version in the meantime.

src/apps/service/apps/de/jobs.clj
dennisroberts71: The extra space has been exterminated. Where's a Dalek emoji when you need one?

src/apps/service/apps/de/jobs.clj
dennisroberts71: Heh, I think I transposed these two lines. I really need to disable that Emacs keybinding. 😆

dennisroberts71:
Thanks for the review!

PR #23, opened 2016-12-14
CORE-8065: fixed a null pointer exception when listing analysis param…

dennisroberts71:
…eters

Fixes an NPE when listing analysis parameters.


dennisroberts71:
This is tiny, so I'm just going to merge it.

PR #24, opened 2016-12-15
CORE-8065: fix a problem with the submission being stored incorrectly…

dennisroberts71:
… when DE jobs are submitted


dennisroberts71:
This change is tiny, so I'm just going to merge it.

PR #25, opened 2016-12-16
Add the system ID path parameter to admin apps endpoints.

dennisroberts71:


src/apps/clients/metadata.clj
```diff
@@ -27,10 +27,11 @@
...
   (let [version (db-categories/get-active-hierarchy-version)]
-    (when (and validate (empty? version))
-      (throw+ {:type  :clojure-commons.exception/not-found
-               :error "An app hierarchy version has not been set."}))
-    version))
+    (if (empty? version)
+      (when validate
+        (throw+ {:type  :clojure-commons.exception/not-found
+                 :error "An app hierarchy version has not been set."}))
+      version)))
```
dennisroberts71: This change ensures that `nil` will be returned whenever the active app hierarchy version has not ben set, which makes it easier for `filter-targets-by-ontology-search` to avoid querying the metadata service if there's no active ontology.

src/apps/clients/metadata.clj
```diff
@@ -54,14 +55,16 @@
...
+  [username category-attrs search-term app-ids & {:keys [validate :or {validate true}]}]
+  (if-let [active-hierarchy-version (get-active-hierarchy-version :validate validate)]
+    (metadata-client/filter-targets-by-ontology-search (config/metadata-client)
+                                                       username
+                                                       active-hierarchy-version
+                                                       category-attrs
+                                                       search-term
+                                                       [app-target-type]
+                                                       app-ids)
+    []))
```
dennisroberts71: This change allows the app search service to work even if there's no active ontology set in the database, which prevents us from having to load an ontology specifically for apps integration tests.

test/apps/test_fixtures.clj
```diff
@@ -31,33 +30,6 @@
...
-                :follow-redirects false}))
-
-(defn with-ontology [f]
-  (let [username         "ipctest"
-        ontology-version (load-ontology username)]
-    (cp/add-hierarchy-version username ontology-version)
-    (f)
-    (delete :app_hierarchy_version (where {:version ontology-version}))
-    (delete-ontology username ontology-version)))
-
```
dennisroberts71: Loading the ontology in a test fixture turned out to be extremely brittle, so I removed the need for it.

src/apps/service/apps/agave.clj
```diff
@@ -286,116 +217,96 @@
...
+
+  (adminDeleteApp [_ system-id app-id]
+    (validate-system-id system-id)
+    (reject-app-integration-request))
+
+  (adminUpdateApp [_ system-id app-id]
+    (validate-system-id system-id)
+    (reject-app-integration-request))
+
+  (getAdminAppCategories [_ _]
```
psarando: /bow
m(_ _)m

PR #26, opened 2017-01-12
CORE-8524: fixed a NullPointerException that was occuring when sharin…

dennisroberts71:
…g analyses


dennisroberts71:
Thanks for the review!

PR #27, opened 2017-01-24
CORE-8540 Allow jobs without outputs to be (un)shared.

psarando:
The `POST /analyses/sharing` and `POST /analyses/unsharing` endpoints have been updated so that they no longer return an error when the results folder can't be (un)shared. Instead, the response will return success, but will now include an `outputs_error` field with an error message for the results (un)share error.

For example:
```json
{
  "unsharing": [
    {
      "user": "username",
      "analyses": [
        {
          "analysis_id": "714d5da0-e1bf-11e6-bf3b-f64e9b87c109",
          "analysis_name": "Word_Count_analysis1",
          "success": true,
          "outputs_error": "unable to unshare result folder: ERR_DOES_NOT_EXIST"
        }
      ]
    }
  ]
}
```

This allows jobs to be (un)shared, even when the job has not yet completed and the output folder doesn't exist, or if the output folder has been renamed.

### DIFF
--- a/src/apps/routes/schemas/permission.clj
+++ b/src/apps/routes/schemas/permission.clj
@@ -82,9 +82,10 @@
 
 (defschema AnalysisSharingResponseElement
   (assoc AnalysisSharingRequestElement
-    :analysis_name        (describe NonBlankString "The analysis name")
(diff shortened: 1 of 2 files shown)


psarando:
Thanks for the review 👍 
I'll merge this after the release.

PR #28, opened 2017-02-03
CORE-8555: Allow sharing of Agave apps and analyses

psarando:
These changes add support for Agave app sharing using the new features added by cyverse-de/mescal#4.
Agave app sharing is the last feature required to enable job sharing for Agave apps.


psarando:
Thanks for the review! 👍

PR #29, opened 2017-02-08
CORE-8593 Update `GET /analyses` can_share flag with permission check.

psarando:
The `can_share` flag will now only be set to `true` if the requesting user also has `own` permission for the job.

A side effect of this change is that the return values of `apps.service.apps.job-listings/list-job` and `apps.service.apps.jobs/update-job-status` will currently always have this flag set to `false`, since they do not lookup permissions, but this `can_share` flag is currently only used in the `GET /analyses` response.

### DIFF
--- a/src/apps/service/apps/job_listings.clj
+++ b/src/apps/service/apps/job_listings.clj
@@ -39,13 +39,15 @@
                   :total (count children)))))
 
 (defn- job-supports-sharing?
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
Looks straightforward to me, I'd say :shipit:

psarando:
Thanks for the review 👍

PR #31, opened 2017-02-24
CORE-8606: added a new endpoint that can be used to obtain informatio…

dennisroberts71:
…n about available system IDs

### DIFF
--- a/src/apps/routes/apps.clj
+++ b/src/apps/routes/apps.clj
@@ -77,17 +77,6 @@
     operation."
     (ok (apps/unshare-apps current-user (:unsharing body))))
 
-  (context "/systems" []
-
-    (GET "/ids" []
-      :query [params SecuredQueryParams]
-      :return SystemIdsResponse
-      :summary "List execution system IDs."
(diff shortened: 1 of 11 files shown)


src/apps/routes/apps.clj
```diff
@@ -77,6 +77,17 @@
     operation."
     (ok (apps/unshare-apps current-user (:unsharing body))))
 
+  (context "/systems" []
+
+    (GET "/ids" []
+      :query [params SecuredQueryParams]
```
psarando: I wonder if this endpoint should be displayed under a different grouping in the Swagger docs?
The `App endpoints` group might be getting cluttered...
Maybe if we rename the `Service Status Information` group, this endpoint could go in that group?

dennisroberts71: Hmm, yeah, the `App endpoints` group might be getting a little bit cluttered. I might create a separate root context instead, though. Reusing the `Service Status Information` group  would work, although I'd either have to modify the endpoint so that it doesn't require user information or add the middleware to the context inside `apps.routes.status`.

dennisroberts71:
The requested change has been made.

PR #32, opened 2017-02-27
CORE-8605: added some validation code back to the app categorization …

dennisroberts71:
…endpoint


dennisroberts71:
Thanks for the review!

PR #34, opened 2017-02-27
CORE-8607: changed the order of evaluation for apps endpoints

dennisroberts71:


dennisroberts71:
Thanks for the review!

PR #35, opened 2017-02-28
CORE-8610: use the last updated time from the Agave job status update…

dennisroberts71:
These changes fix two problems.

The first problem is that Agave jobs are not updated with an end timestamp when the job moves to either the `Completed` state or the `Failed` state (CORE-8610). This was happening because the end date was not being sent to the DE in the job status updates. To fix this problem, I modified the services so that they will use the `lastUpdated` timestamp if the end time is not provided in the status update.

The second problem is that the services aren't verifying app permissions in all cases. This wasn't a problem before because app listings were checking permissions. Now that it's possible to share an analysis with a user without the corresponding app being shared, however, it is possible for users to run apps without having explicit permission to. I modified the services to check permissions on an app any time a user tries to obtain its job submission information.

### DIFF
--- a/src/apps/routes/callbacks.clj
+++ b/src/apps/routes/callbacks.clj
@@ -7,7 +7,7 @@
 
 (defroutes callbacks
   (POST "/de-job" []
-         :body [body (describe DeJobStatusUpdate "The App to add.")]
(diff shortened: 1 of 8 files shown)


dennisroberts71:
Thanks for the review!

PR #38, opened 2017-03-09
CORE-8595: allow administrative app deletions to succeed if the app i…

dennisroberts71:
…s not defined in the permissions database

The endpoint to delete a resource will only return a 404 if the resource doesn't exist. If the app exists in the DE database but the corresponding resource doesn't exist in the permissions database then the databases are out of sync and DE administrators should still be permitted to delete the app.


dennisroberts71:
Thanks for the review! 👍

PR #39, opened 2017-03-15
CORE-8418: Add permissions support for public tools

psarando:
This PR updates the following endpoints with public permissions support for tools (all tools are currently considered public):

* `POST /admin/tools`
* `DELETE /admin/tools/:id`
* `GET /tools`

A future PR will add support to allow users to add private tools, and to restrict tool listing and details based on private vs. public permissions. This PR is just to ensure that tool permissions stay in sync with any tools added or deleted by admins before any further features are implemented.

### DIFF
--- a/src/apps/tools.clj
+++ b/src/apps/tools.clj
@@ -8,7 +8,8 @@
         [kameleon.queries]
         [kameleon.util.search]
         [korma.core :exclude [update]]
-        [korma.db :only [transaction]])
+        [korma.db :only [transaction]]
+        [slingshot.slingshot :only [try+]])
   (:require [apps.clients.permissions :as permissions]
             [apps.persistence.app-metadata :as persistence]
             [clojure.tools.logging :as log]))
@@ -113,4 +114,8 @@
     (validate-tool-not-used tool-id)
     (log/warn user "deleting tool" tool-id name version "@" location))
   (delete tools (where {:id tool-id}))
+  (try+
+    (permissions/delete-tool-resource tool-id)
+    (catch [:status 404] _
+      (log/warn "tool resource" tool-id "not found by permissions service")))
   nil)
--- a/src/apps/tools.clj
+++ b/src/apps/tools.clj
@@ -9,7 +9,8 @@
         [kameleon.util.search]
         [korma.core :exclude [update]]
         [korma.db :only [transaction]])
-  (:require [apps.persistence.app-metadata :as persistence]
+  (:require [apps.clients.permissions :as permissions]
+            [apps.persistence.app-metadata :as persistence]
             [clojure.tools.logging :as log]))
 
 (defn- add-search-where-clauses
@@ -96,6 +97,7 @@
   [{:keys [tools]}]
   (transaction
     (let [tool-ids (doall (map add-new-tool tools))]
+      (dorun (map permissions/register-public-tool tool-ids))
       {:tool_ids tool-ids})))
 
 (defn update-tool
--- a/src/apps/tools.clj
+++ b/src/apps/tools.clj
@@ -60,19 +60,29 @@
          (add-query-offset offset)
          (add-hidden-tool-types-clause include-hidden)))))
 
+(defn format-tool-listing
+  [perms public-tool-ids {:keys [id] :as tool}]
+  (-> tool
+      (assoc :is_public  (contains? public-tool-ids id)
+             :permission (or (perms id) ""))
+      remove-nil-vals))
+
 (defn search-tools
   "Obtains a listing of tools for the tool search service."
-  [params]
-  {:tools
-     (map remove-nil-vals
-       (select (tool-listing-base-query params)))})
+  [{:keys [user] :as params}]
+  (let [perms           (permissions/load-tool-permissions user)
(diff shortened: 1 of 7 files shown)


src/apps/tools.clj
```diff
@@ -96,19 +108,24 @@
   [{:keys [tools]}]
   (transaction
     (let [tool-ids (doall (map add-new-tool tools))]
+      (dorun (map permissions/register-public-tool tool-ids))
```
psarando: Since this is inside a `transaction`, I think this could run the risk of leaving orphaned entries in the permissions service, if something goes wrong while adding these permissions. But at least it won't attempt to add any permissions if something goes wrong while adding one of the tools.

dennisroberts71: Yeah, I ended up doing something similar for apps. It probably is better to err on the side of not saving things in the apps database if the permissions can't be created.

src/apps/tools.clj
```diff
@@ -58,19 +60,29 @@
...
-  [params]
-  {:tools
-     (map remove-nil-vals
-       (select (tool-listing-base-query params)))})
+  [{:keys [user] :as params}]
+  (let [perms           (permissions/load-tool-permissions user)
+        public-tool-ids (permissions/get-public-tool-ids)]
+    {:tools
+     (map (partial format-tool-listing perms public-tool-ids)
+       (select (tool-listing-base-query params)))}))
```
dennisroberts71: I could be missing something here, but it looks as though tools that the user does not have permissions to view may be included in the results.

src/apps/tools.clj
```diff
@@ -58,19 +60,29 @@
...
 
 (defn get-tool
   "Obtains a tool by ID."
-  [tool-id]
+  [user tool-id]
   (let [tool           (->> (first (select (tool-listing-base-query) (where {:tools.id tool-id})))
                             (assert-not-nil [:tool-id tool-id])
-                            remove-nil-vals)
+                            (format-tool-listing (permissions/load-tool-permissions user)
+                                                 (permissions/get-public-tool-ids)))
```
dennisroberts71: I don't see how this function prevents users from viewing tools that they're not permitted to view either.

src/apps/tools.clj
```diff
@@ -96,19 +108,24 @@
...
       {:tool_ids tool-ids})))
 
 (defn update-tool
-  [overwrite-public {:keys [id container] :as tool}]
+  [user overwrite-public {:keys [id container] :as tool}]
   (persistence/update-tool tool)
   (when container
     (set-tool-container id overwrite-public container))
-  (get-tool id))
+  (get-tool user id))
```
dennisroberts71: I don't see where this function is checking tool permissions before allowing the update.

src/apps/tools.clj
```diff
@@ -96,19 +108,24 @@
...
-  (let [{:keys [name location version]} (get-tool tool-id)]
+  (let [{:keys [name location version]} (get-tool user tool-id)]
     (validate-tool-not-used tool-id)
     (log/warn user "deleting tool" tool-id name version "@" location))
   (delete tools (where {:id tool-id}))
+  (try+
+    (permissions/delete-tool-resource tool-id)
+    (catch [:status 404] _
+      (log/warn "tool resource" tool-id "not found by permissions service")))
   nil)
```
dennisroberts71: I don't see where this function is checking tool permissions either.

src/apps/tools.clj
psarando: That's true for now. As the summary mentioned, all tools are currently considered public, and this PR is just to ensure that tool permissions stay in sync with any tools added or deleted by admins before any further features are implemented.
In a future PR, when users are allowed to add private tools, this function will be updated to restrict the results based on permissions.

dennisroberts71: Heh, Derp. Of course, I looked directly at the diff and didn't read the summary.

src/apps/tools.clj
psarando: This is an admin endpoint, and only admins will be allowed to delete public tools.
Once users are allowed to add private tools, I'm not sure if we'll want to allow admins to delete private tools?

dennisroberts71: 😖

dennisroberts71: Heh, sorry. I should have looked at where this was called.

psarando: Ah, no problem. It's not really obvious from inside this namespace 😁 

This also applies for the update endpoint. Once the features are added for private tool support, there will be endpoints specific to private tools that will check permissions.

I think for now I'm leaning towards not allowing admins to update or delete private tools. It shouldn't be too much work to add that support later if we decide we need it.

psarando:
Thanks for the review! 👍

src/apps/tools.clj
dennisroberts71: Yeah, that sounds reasonable. It will be better when we can add the admin permission level. The primary reason that I hadn't added it yet was because we didn't have LDAP groups available to the DE, which meant that permissions couldn't easily be granted to all administrator groups. Now that we have LDAP groups being loaded, we'll probably want to revisit this at some point in the near future.

PR #40, opened 2017-03-30
CORE-8696: use the actual parameter label in the app task listing for…

dennisroberts71:
… output file parameters that have no default output file name


dennisroberts71:
Thanks for the review!

PR #41, opened 2017-03-30
CORE-8691 Add a `POST /tools` endpoint that allows users to add private tools

psarando:
Private tools are currently not allowed to mount any devices or volumes, and have restrictions on networking (none), resources (4 CPUs / 8GB RAM), and run duration (24 hours). If implementation details are not provided in the request, then the requesting user's name and email will be used by default.

This PR also adds permission checks on tool listings and private tool details.
Admin endpoints will not currently list a private tool if the admin does not have permissions on that tool.

### DIFF
--- a/src/apps/routes/schemas/containers.clj
+++ b/src/apps/routes/schemas/containers.clj
@@ -189,13 +189,17 @@
    {:container_volumes_from [VolumesFrom]}
    "The list of VolumeFroms associated with a tool's container."))
 
+(def DevicesParamOptional     (s/optional-key :container_devices))
+(def VolumesParamOptional     (s/optional-key :container_volumes))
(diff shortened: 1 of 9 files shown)


src/apps/routes/schemas/containers.clj
```diff
@@ -189,13 +189,17 @@
    {:container_volumes_from [VolumesFrom]}
    "The list of VolumeFroms associated with a tool's container."))
 
+(def DevicesParamOptional     (s/optional-key :container_devices))
+(def VolumesParamOptional     (s/optional-key :container_volumes))
+(def VolumesFromParamOptional (s/optional-key :container_volumes_from))
```
dennisroberts71: 👍

PR #42, opened 2017-04-07
CORE-7507 Update the apps is-publishable endpoint to check for app existence first

psarando:
Now that this endpoint is guarded with a `system-id`, the apps service can check if a DE app exists first, and return an `ERR_NOT_FOUND` before checking the user's app permissions.

This PR also removes some obsolete code.


psarando:
Thanks for the review 👍

PR #43, opened 2017-04-11
CORE-8692: Add endpoints that allow users to (un)share private tools

psarando:
This PR adds `POST /tools/sharing` and `POST /tools/unsharing` endpoints.
These endpoints will send tool (un)sharing notifications to the sharer and sharee as appropriate.

Also, when a user shares an app that uses private tools, those tools are also shared with `read` access.


psarando:
Thanks for the review! 👍

PR #45, opened 2017-04-18
CORE-8693 Add `PATCH /tool/:tool-id` endpoint.

psarando:
This PR adds a `PATCH /tool/:tool-id` endpoint to allow users to update their private tools.

As with the `POST /tools` endpoint, `type` is always set to `executable` and `restricted` is always set to `true`, even if other values are set in the request, and a configured limit may override the `time_limit_seconds` field set in the request.

Also, similar to the `PATCH /admin/tool/:tool-id` endpoint, if the `container` object is omitted in the request, then existing container settings will not be modified, but if the `container` object is present in the request, then all container settings must be included in it. Any existing settings not included in the request's `container` object will be removed, except `network_mode` is always set to `none` and configured limits may override values set (or omitted) for the `cpu_shares` and `memory_limit` fields.

The `POST /tools` endpoint was also updated to allow `implementation` details in its requests (also allowed in the new PATCH endpoint).

### DIFF
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -14,7 +14,7 @@
         [apps.util.conversions :only [remove-nil-vals remove-empty-vals]]
         [apps.validation :only [validate-image-not-public
                                 validate-image-not-used
-                                validate-tool-not-public]]
+                                validate-tool-not-used-in-public-apps]]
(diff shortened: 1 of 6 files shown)


psarando:
Thanks for the review 👍

PR #46, opened 2017-04-19
CORE-8694 Add DELETE /tools/:tool-id endpoint.

psarando:
This PR adds a `DELETE /tools/:tool-id` endpoint that allows a user to delete a private tool, as long as it is not in use by any apps.

The requesting user must have ownership permission for the tool.

If the tool is already in use in private apps, then an `ERR_NOT_WRITEABLE` will be returned along with a listing of the apps using this tool (same as the `DELETE /admin/tools/:tool-id` endpoint); unless the `force-delete` flag is set to `true`, then the tool will be removed from those apps and then deleted.

### DIFF
--- a/src/apps/routes/tools.clj
+++ b/src/apps/routes/tools.clj
@@ -7,8 +7,8 @@
         [apps.routes.schemas.containers]
         [apps.routes.schemas.integration-data :only [IntegrationData]]
         [apps.routes.schemas.tool]
-        [apps.tools :only [add-tools delete-tool get-tool search-tools update-tool]]
-        [apps.tools.private :only [add-private-tool update-private-tool]]
+        [apps.tools :only [add-tools admin-delete-tool get-tool search-tools update-tool]]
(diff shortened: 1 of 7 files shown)


src/apps/routes/tools.clj
```diff
@@ -182,6 +182,25 @@ otherwise the default value will be used."
...
+          :path-params [tool-id :- ToolIdParam]
+          :query [{:keys [user force-delete]} PrivateToolDeleteParams]
+          :responses (merge CommonResponses
+                            {200 {:description "The Tool was successfully deleted."}
+                             400 {:schema      ErrorResponseNotWritable
+                                  :description "The Tool could not be deleted."}
+                             403 {:schema      ErrorResponseForbidden
+                                  :description "The requesting user does not have permission to delete this Tool."}
+                             404 {:schema      ErrorResponseNotFound
+                                  :description "A Tool with the given `tool-id` does not exist."}})
```
dennisroberts71: 👍

psarando:
Thanks for the review 👍

PR #47, opened 2017-04-24
CORE-8748 / CORE-7531: Update tool listings and search

psarando:
This PR updates the `GET /tools` listing endpoint (which also updates the `GET /apps/elements/tools` listing response):

* The `search` param is now optional.
* An optional `public` param now allows `GET /tools` to return only public, only private, or all tools.
* The tool's Docker image `name` and `tag` are now included in listing results.
* The tool's implementor name and implementor email are now included in listing results.
* The `search` param can now match a tool's Docker image name, integrator name, or integrator email.

### DIFF
--- a/src/apps/persistence/tools.clj
+++ b/src/apps/persistence/tools.clj
@@ -133,34 +133,37 @@
     base-query))
 
 (defn- tool-listing-base-query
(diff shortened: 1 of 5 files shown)


psarando:
Thanks for the review 👍

PR #48, opened 2017-04-26
CORE-8697 Add `GET /admin/tools` and `GET /admin/tools/:tool-id` endpoints

psarando:
This PR adds admin endpoints to list all tools and fetch details for any tool.

The swagger docs for some /admin/tools endpoints were also updated, and the `PATCH /admin/tools/:id` endpoint now validates a tool exists before attempting to update it.


psarando:
Thanks for the review! 👍

PR #49, opened 2017-04-28
CORE-8745 Make tool version required and location optional.

psarando:
This PR allows the `location` field to be optional in `tools` endpoints, since all new tools are using Docker images and the `location` field is no longer required for running new tools.

Since we have many tools with duplicate names, this PR also makes the `version` field required (also updated by cyverse-de/de-db#8).


psarando:
Cool, thanks for the review! 👍

PR #50, opened 2017-04-29
CORE-7532: modified the app details endpoint to return tool container…

dennisroberts71:
… information rather than tool location information for DE apps

### DIFF
--- a/src/apps/routes/schemas/app.clj
+++ b/src/apps/routes/schemas/app.clj
@@ -7,6 +7,7 @@
                                           SortFieldOptionalKey]]
         [apps.routes.params]
         [apps.routes.schemas.app.rating]
+        [apps.routes.schemas.containers :as containers]
         [apps.routes.schemas.tool :only [Tool]]
         [schema.core :only [Any defschema enum optional-key recursive]])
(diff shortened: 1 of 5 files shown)


src/apps/routes/schemas/app.clj
```diff
@@ -261,8 +262,9 @@
    :name (describe String "The App Category's name")})
 
 (defschema AppDetailsTool
-  (assoc Tool
-    :id (describe String "The tool identifier.")))
+  (merge Tool
+         {:id                       (describe String "The tool identifier.")
+          (optional-key :container) {:image (dissoc containers/Image :id)}}))
```
psarando: It looks like the same `:image` schema is used in the [ToolListingItem](https://github.com/cyverse-de/apps/blob/master/src/apps/routes/schemas/tool.clj#L65).
Maybe this could be refactored into a common schema in `apps.routes.schemas.tool`?

It's also pretty simple and only used in 2 places so far, so I don't think it should be required for this PR.

dennisroberts71: Thanks for the suggestion. I think I'll probably make it before merging this request just so that I don't forget to do it later.

PR #51, opened 2017-05-02
CORE-8383: convert the apps service to Consul configs

dennisroberts71:


dennisroberts71:
Thanks for the review! Making the default properties optional sounds reasonable to me. Thankfully, my editor has macros. 😆

PR #52, opened 2017-05-05
CORE-8798 Add private tool validation to is-publishable and publish endpoints

psarando:
Apps using private tools will now cause `GET /apps/:system-id/:app-id/is-publishable` to return false and `POST /apps/:system-id/:app-id/publish` to return an error.


psarando:
Thanks for the review! 👍

PR #53, opened 2017-05-05
CORE-8811 Update Tool schema in apps add/edit and arg-preview endpoints.

psarando:
Allows the UI to pass extra tool listing fields in the request body of the apps add, edit, and arg-preview endpoints, since it's not filtering those tool fields it gets from the listing endpoint when submitting the tools with an app request.

The the response of the `PUT /apps/{system-id}/{app-id}` endpoint also needed to be updated to the same response as the `GET /apps/{system-id}/{app-id}/ui` endpoint in order to avoid returning these extra tool fields received from the UI.


psarando:
Thanks for the review! 👍

PR #54, opened 2017-05-05
CORE-8746 Allow tool request `architecture` field to be optional.

psarando:
The `POST /tool-requests` endpoint no longer requires the `architecture` field.

Since a `tool_architecture_id` is still required by the `tool_requests` db table, the service will default this value to `Others` if it's not sent in the request from the client.


psarando:
Thanks for the review 👍

PR #56, opened 2017-05-09
CORE-8814 Add GET (/admin)/tools/:tool-id/apps endpoints.

psarando:
Adds endpoints for listing apps using a given tool.
The non-admin endpoint also checks if the requesting user has `read` permission for the given tool.

### DIFF
--- a/src/apps/routes/tools.clj
+++ b/src/apps/routes/tools.clj
@@ -6,6 +6,7 @@
         [apps.routes.params]
         [apps.routes.schemas.containers]
         [apps.routes.schemas.integration-data :only [IntegrationData]]
+        [apps.routes.schemas.app :only [AdminAppListing AppListing]]
         [apps.routes.schemas.tool]
         [apps.tools :only [admin-add-tools
                            admin-delete-tool
@@ -16,11 +17,13 @@
                            user-get-tool]]
         [apps.tools.private :only [add-private-tool delete-private-tool update-private-tool]]
         [apps.user :only [current-user]]
+        [apps.util.coercions :only [coerce!]]
         [apps.util.service]
         [slingshot.slingshot :only [throw+]]
         [ring.util.http-response :only [ok]])
   (:require [apps.routes.schemas.permission :as permission]
             [apps.service.apps :as apps]
+            [apps.service.apps.de.listings :as app-listings]
             [apps.tools.permissions :as tool-permissions]
             [apps.tools.sharing :as tool-sharing]
             [compojure.api.middleware :as middleware]))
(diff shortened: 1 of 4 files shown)


src/apps/routes/tools.clj
```diff
@@ -16,11 +17,13 @@
                            user-get-tool]]
         [apps.tools.private :only [add-private-tool delete-private-tool update-private-tool]]
         [apps.user :only [current-user]]
+        [apps.util.coercions :only [coerce!]]
         [apps.util.service]
         [slingshot.slingshot :only [throw+]]
         [ring.util.http-response :only [ok]])
   (:require [apps.routes.schemas.permission :as permission]
             [apps.service.apps :as apps]
+            [apps.service.apps.de.listings :as app-listings]
```
dennisroberts71: This would have to be done as part of a larger refactor, but I'd eventually like to have all app listing operations go directly through `apps.service.apps`.

src/apps/service/apps/de/listings.clj
```diff
@@ -290,19 +291,41 @@
...
         app-ids         (set (keys perms))
-        app-listing-ids (metadata-filter app-ids)
-        beta-ids-set    (app-ids->beta-ids-set shortUsername app-listing-ids)
-        public-app-ids  (perms-client/get-public-app-ids)
-        app-listing-fn  (if admin? admin-list-apps-by-id list-apps-by-id)
-        app-listing     (app-listing-fn workspace faves-index app-listing-ids (fix-sort-params params))]
-    {:total (count app-listing-ids)
-     :apps  (map (partial format-app-listing admin? perms beta-ids-set public-app-ids) app-listing)}))
+        app-listing-ids (metadata-filter app-ids)]
+    (app-listing-by-id user params perms app-listing-ids admin?)))
```
dennisroberts71: I like the refactoring that you did here. 👍

src/apps/routes/tools.clj
psarando: I was wondering about that. I thought that since tools were specific to the DE (we don't support tools in Agave endpoints) that we didn't need to route anything tool-specific through that interface.

If we might support tools in future 3rd-party services, then this makes sense.

dennisroberts71: Yeah, Agave doesn't have the concept of a tool, which makes this unnecessary for now. I'm not sure whether or not we'll ever integrate a third-party execution platform that supports separate tools and apps, so it may never be necessary from a functional standpoint.

The other reason I'd like to have all requests go through that interface is kind of aesthetic. Limiting calls to that namespace hierarchy makes it slightly easier to refactor the code in the future if we need to.

This isn't a huge deal, and certainly isn't something that we need to tackle now; it's just something that I'd like to do sometime.

psarando:
Cool, thanks for the review 👍

PR #57, opened 2017-05-12
Added the subject source ID to sharing endpoints.

dennisroberts71:
The app, tool, and analysis sharing endpoints are all affected.

I'm going to be testing this and probably making some changes as I find problems, but I wanted to submit this PR as soon as I could so that I could hopefully get another set of eyes on it.


dennisroberts71:
Some simple manual tests have been done for this change, so I'm going to merge this request immediately. The service currently sends a sharing notification to the sharer for each member of the group, which could be overwhelming. I've filed an issue suggesting that this should be changed.

PR #59, opened 2017-05-18
CORE-8750: Add `deprecated` flag to tool images

psarando:
This PR adds the `deprecated` flag to tool details in app and tool listing endpoints.

The following endpoints now use the `deprecated` flag in the tool JSON object:
* `PUT /apps/{system-id}/{app-id}`
* `POST /apps/{system-id}/{app-id}/copy`
* `GET /apps/{system-id}/{app-id}/ui`

The following endpoints now use the `deprecated` flag in the tool's container image JSON object:
* `GET /admin/apps/{system-id}/{app-id}/details`
* `GET /admin/tools`
* `POST /admin/tools`
* `GET /admin/tools/container-images`
* `POST /admin/tools/container-images`
* `GET /admin/tools/container-images/{image-id}`
* `PATCH /admin/tools/container-images/{image-id}`
* `GET /admin/tools/{tool-id}`
* `PATCH /admin/tools/{tool-id}`
* `GET /apps/{system-id}/{app-id}/details`
* `GET /tools`
* `POST /tools`
* `GET /tools/{tool-id}`
* `PATCH /tools/{tool-id}`

Tools using deprecated images will no longer be included in user tool listings.
Admins can update container image `deprecated` flags with the `PATCH /admin/tools/container-images/{image-id}` endpoint.
The `POST /tools` and `PATCH /tools/{tool-id}` endpoints will return an `ERR_BAD_OR_MISSING_FIELD` if the request includes a container image that has been deprecated.
The `POST /apps/{system-id}/{app-id}/publish` endpoint will return an `ERR_BAD_OR_MISSING_FIELD` if the app uses a deprecated tool (if the tool is using a deprecated image).
The `POST /tool-requests` endpoint will return an `ERR_BAD_OR_MISSING_FIELD` if the request includes a `tool_id` for a deprecated tool.

### DIFF
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -105,7 +105,7 @@
   "Updates the record for a container image. Basically, just allows you to set a new URL
    at this point."
   [image-id user overwrite-public image-info]
-  (let [update-info (select-keys image-info [:name :tag :url])]
+  (let [update-info (select-keys image-info [:name :tag :url :deprecated])]
     (when-not (empty? update-info)
       (when-not overwrite-public
         (validate-image-not-public image-id))
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -81,11 +81,11 @@
 (defn find-or-add-image-info
   "Finds or inserts an image with the given name and tag in the container_images table."
   [{:keys [name tag url] :or {tag "latest"} :as image-map}]
-  (let [existing-image (find-image-by-name-and-tag image-map)]
-    (or existing-image
-        (insert container-images (values {:name name
-                                          :tag  tag
-                                          :url  url})))))
+  (if-let [existing-image (find-image-by-name-and-tag image-map)]
+    existing-image
+    (insert container-images (values {:name name
+                                      :tag  tag
+                                      :url  url}))))
 
 (defn image?
   "Returns true if the given name and tag exist in the container_images table."
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -47,11 +47,11 @@
     (when registry
       (encode-auth registry))))
 
-(defn public-image-info
+(defn- public-image-info
   "Returns a map containing only publicly-permissible image info (no auth)"
   [image-uuid]
   (first (select container-images
-                (fields :name :tag :url :id)
+                (fields :name :tag :url :deprecated :id)
                 (where {:id (uuidify image-uuid)}))))
 
 (defn image-info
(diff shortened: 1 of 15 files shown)


src/apps/tools/private.clj
```diff
@@ -9,6 +10,14 @@
             [apps.util.config :as cfg]
             [apps.validation :as validation]))
 
+(defn- validate-image-not-deprecated
+  [image-info]
+  (let [image (containers/find-image-by-name-and-tag image-info)]
+    (when (:deprecated image)
+    (throw+ {:type  :clojure-commons.exception/bad-request-field
+             :error "Image is deprecated and should not be used in new tools."
+             :image image}))))
```
dennisroberts71: This `throw+` call might be another good candidate for `clojure-commons.exception-util`. It doesn't save that much space, but I like to use those functions for convenience just because I'm lazy and it saves me typing. 😆

src/apps/validation.clj
```diff
@@ -184,11 +184,13 @@
...
   [tool]
-  (let [existing-tool (first (select tools (where (select-keys tool [:name :version]))))]
+  (let [existing-tool (first (select tools
+                                     (fields :id :name :version)
+                                     (where (select-keys tool [:name :version]))))]
     (when existing-tool
       (throw+ {:type  :clojure-commons.exception/exists
                :error "A Tool with that name and version already exists."
-               :tool  tool}))))
+               :tool  existing-tool}))))
```
dennisroberts71: There actually is a `clojure-commons.exception-util` function for this one. You could do something like this (assuming the namespace alias is `cxu`):

```clojure
(cxu/exists "A Tool with that name and version already exists." :tool existing-tool)
```

psarando: Cool, thanks for the suggestion 👍  I'll look into refactoring this to use the alternate function.

src/apps/containers.clj
```diff
@@ -84,54 +72,40 @@
...
+(defn find-image-by-name-and-tag
+  "Finds the image matching the given name and tag in the container_images table."
+  [{:keys [name tag] :or {tag "latest"}}]
+  (first (select container-images (where {:name name
+                                          :tag  tag}))))
+
+(defn find-or-add-image-info
+  "Finds or inserts an image with the given name and tag in the container_images table."
+  [{:keys [name tag url] :or {tag "latest"} :as image-map}]
+  (let [existing-image (find-image-by-name-and-tag image-map)]
```
psarando: I was thinking about cleaning this function up.
I like your suggestion, so I'll refactor this before I merge this PR 👍

psarando:
Thanks for the review! 👍

PR #60, opened 2017-05-18
CORE-8834: modified the permission-lister endpoints for apps, tools, …

dennisroberts71:
…and analyses to include the subject source ID in the response body

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -24,7 +24,7 @@
                  [org.cyverse/authy "2.8.0"]
                  [org.cyverse/clojure-commons "2.8.3"]
                  [org.cyverse/kameleon "3.0.1"]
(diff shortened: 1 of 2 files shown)


project.clj
```diff
@@ -24,7 +24,7 @@
                  [org.cyverse/authy "2.8.0"]
                  [org.cyverse/clojure-commons "2.8.3"]
                  [org.cyverse/kameleon "3.0.1"]
-                 [org.cyverse/mescal "2.8.5"]
+                 [org.cyverse/mescal "2.8.6-SNAPSHOT"]
```
psarando: We can cut a release of `mescal 2.8.6` now, and use that in this PR.

dennisroberts71:
Thanks for the review! 👍

PR #61, opened 2017-05-24
CORE-8857 Add tool details to app tasks and pipeline ui endpoints.

psarando:
This PR will add tool details to each DE task in the following endpoint responses:
* `GET /apps/de/{app-id}/tasks`
* `GET /apps/pipelines/{app-id}/ui`

This will allow the UI to check for deprecated tools in apps added to workflows, or already included in workflows when first opened for editing.


psarando:
Thanks for the review 👍

PR #62, opened 2017-05-26
CORE-8695 Add POST /admin/tools/{tool-id}/publish

psarando:
Adds an endpoint for admins to make tools public, also allowing the admin to make updates to the tool in the same request.


psarando:
Thanks for the review 👍

PR #63, opened 2017-06-01
CORE-8747 Add tool request ID and status to tool listings.

psarando:
This PR adds tool request info like the following to tool listings, if the tool has a tool request associated with it:
```json
{
  "tool_request": {
    "id": "uuid",
    "status": "string"
  }
}
```
The following endpoints have been updated with this response info:
* `GET /admin/tools`
* `GET /apps/elements/tools`
* `GET /tools`

### DIFF
--- a/src/apps/persistence/tools.clj
+++ b/src/apps/persistence/tools.clj
@@ -162,9 +162,11 @@
 (defn- tool-request-status-subselect
   []
   (subselect [:tool_request_status_codes :trsc]
-             (fields [(sqlfn :last :trsc.name) :status])
+             (fields [:trsc.name :status])
              (join [:tool_request_statuses :trs] {:trs.tool_request_status_code_id :trsc.id})
-             (where {:trs.tool_request_id :tool_requests.id})))
+             (where {:trs.tool_request_id :tool_requests.id})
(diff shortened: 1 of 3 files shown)


src/apps/persistence/tools.clj
```diff
@@ -159,6 +159,13 @@
               [:tools.time_limit_seconds :time_limit_seconds])
       (join tool_types)))
 
+(defn- tool-request-status-subselect
+  []
+  (subselect [:tool_request_status_codes :trsc]
+             (fields [(sqlfn :last :trsc.name) :status])
+             (join [:tool_request_statuses :trs] {:trs.tool_request_status_code_id :trsc.id})
+             (where {:trs.tool_request_id :tool_requests.id})))
```
dennisroberts71: With the `last` aggregation function, it might be a good idea to put an explicit order on this subquery in case the rows aren't returned in chronological order. Since you're using a subquery, you could also avoid using the aggregation function altogether by doing a descending sort on `date_assigned` and limiting the result set to one row.

psarando: Thanks, that's a much better suggestion. 👍

psarando:
Thanks for the review! 👍

PR #66, opened 2017-06-14
CORE-8895: update the way that the "shared with me" app listings are …

dennisroberts71:
…generated to work with group sharing


dennisroberts71:
Thanks for the review! 👍

PR #67, opened 2017-06-14
CORE-8885 Replace private tools cpu_shares limit with pids_limit.

psarando:
This PR removes the `cpu_shares` limit automatically added to private tools, replacing it with initial support for `pids_limit` restrictions in the `POST /tools` endpoint.

The `pids_limit` is only added in this PR as a config setting (64 by default) and as an allowed `container` field in tool requests, but since it's not persisted in the database yet, it won't be returned in tool info or submitted with job info.

### DIFF
--- a/src/apps/routes/schemas/containers.clj
+++ b/src/apps/routes/schemas/containers.clj
@@ -39,6 +39,7 @@
 (s/defschema Settings
   (describe
    {(s/optional-key :cpu_shares)         Integer
+    (s/optional-key :pids_limit)         Integer
     (s/optional-key :memory_limit)       Long
     (s/optional-key :min_memory_limit)   Long
     (s/optional-key :min_cpu_cores)      Integer
--- a/src/apps/util/config.clj
+++ b/src/apps/util/config.clj
@@ -102,10 +102,10 @@
   [props config-valid configs]
   "apps.tools.private.time-limit-seconds" (* 24 60 60)) ;; 24 hours
 
-(cc/defprop-optint private-tool-cpu-shares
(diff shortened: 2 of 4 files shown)


src/apps/util/config.clj
```diff
@@ -102,10 +102,10 @@
   [props config-valid configs]
   "apps.tools.private.time-limit-seconds" (* 24 60 60)) ;; 24 hours
 
-(cc/defprop-optint private-tool-cpu-shares
```
ianmcorvidae: I guess the rest aren't either, but I guess this stuff isn't in the `apps.properties.tmpl`? Not sure if we'll need it to be or if we're just worrying about that if/when we do actually need it.

src/apps/routes/schemas/containers.clj
```diff
@@ -39,6 +39,7 @@
 (s/defschema Settings
   (describe
    {(s/optional-key :cpu_shares)         Integer
```
ianmcorvidae: Should this be removed as well?

psarando: Nope, we want to leave `cpu_shares` as an optional restriction admins can still set.
This PR is just removing the auto-restriction set on private tools added by users.

src/apps/util/config.clj
psarando: Right, these weren't added to the config repos since they're intended as service defaults (and if needed the service can use a different value by updating the live config file).

I'm not sure we need them in the config files yet (we don't need different values in different environments).

psarando:
Thanks for the reviews! 👍

PR #68, opened 2017-06-17
CORE-8885 Update (/admin)/tools endpoints to persist pids_limit setting.

psarando:
The `POST (/admin)/tools` and `PATCH (/admin)/tools/:id` endpoints will now save `pids_limit` container settings to the db. This setting will also now be included in `GET (/admin)/tool/:id` responses and in job submissions.

Note that this PR depends on cyverse-de/de-db#13 getting merged first.

### DIFF
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -367,7 +367,8 @@
 (defn- filter-container-settings
   [settings-map]
   (select-keys settings-map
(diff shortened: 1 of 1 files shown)


psarando:
This is a small change, so I'm going to go ahead and merge it now before today's QA drop.
Reviews are still welcome.

PR #69, opened 2017-06-20
CORE-8916: added an admin endpoint to delete a user's OAuth token

dennisroberts71:


dennisroberts71:
Sriram asked about the possibility of eventually allowing the user to click a button in the UI to delete their own access token, so I added a non-admin endpoint for that purpose.

dennisroberts71:
Thanks for the review! 👍

PR #70, opened 2017-06-22
CORE-7290 Fix DELETE/PATCH /admin/tools/container-images/:id validation.

psarando:
Updated the `DELETE` and `PATCH /admin/tools/container-images/:id` endpoint validations to look for all tools using the given image, not just for tools already used in apps, and also to check for tools using the given image as a data container.

Also renamed some functions related to images in public apps for clarity, since tools can be private now.

### DIFF
--- a/src/apps/persistence/app_metadata.clj
+++ b/src/apps/persistence/app_metadata.clj
@@ -266,13 +266,6 @@
                   [:container_images.deprecated :deprecated])
           (where {:app_id app-id})))
 
-(defn get-tools-by-image-id
-  "Loads information about the tools associated with a Docker image."
-  [image-id]
-  (select (get-tool-listing-base-query)
-          (modifier "DISTINCT")
-          (where {:container_images_id image-id})))
-
 (defn get-public-tools-by-image-id
   "Loads information about public tools associated with a Docker image."
   [img-id]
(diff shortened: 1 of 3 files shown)


src/apps/persistence/app_metadata.clj
```diff
@@ -266,20 +266,27 @@
...
+(defn get-tools-in-public-apps-by-image-id
+  "Loads information about tools used by public apps associated with a Docker image."
   [img-id]
   (select (get-tool-listing-base-query)
           (modifier "DISTINCT")
-          (where {:container_images_id img-id
-                  :app_id              [in (perms-client/get-public-app-ids)]})))
+          (where (or {:app_id              [in (perms-client/get-public-app-ids)]
+                      :container_images_id img-id}
+                     {:tool_id [in (subselect-tool-ids-using-data-container img-id)]}))))
```
dennisroberts71: It appears that this query will also match containers that are used as data containers in exclusively private tools. Should this query be modified so that it only matches containers that are either used in a tool that is used in a public app or used as a data container in at least one tool that is used as a public app?

psarando: Good catch 👍 
I think I can update the query to always match tools with public app IDs, but match the image ID in the tool's `container_images_id` OR in the subselect.

dennisroberts71: Yeah, I think that should work. 👍

psarando:
Thanks for the review! 👍

PR #71, opened 2017-06-23
CORE-8921 Add ImagePublicAppToolListing response schema.

psarando:
Added a `GET /admin/tools/container-images/{image-id}/public-tools` listing response schema with fewer details than the schema used by `GET (/admin)/tools`, since this endpoint does not use the tool listing and formatting functions that include the extra details required for the `(/admin)/tools` listing endpoints.

This fixes response schema validation errors for images used by public apps.


psarando:
This is such a small change that I'm going to go ahead and merge it, but reviews are still welcome.

PR #72, opened 2017-07-11
CORE-7529: ignore inputs for which a blank path has been specified

dennisroberts71:


dennisroberts71:
That seems reasonable, I'll modify it to use `fs/absolute?`.

dennisroberts71:
Thanks for the review!

dennisroberts71:
Thanks for the review!

PR #73, opened 2017-07-25
Allow using the path-info endpoint, add filtering.

ianmcorvidae:
As far as I can tell these are the things which are needed in each of these calling locations. `lein check` and `lein eastwood` and `lein kibit` are fine with this code. I've dealt with every location that previously called `apps.clients.data-info/get-file-stats`, since I changed the name.

### DIFF
--- a/src/apps/clients/data_info.clj
+++ b/src/apps/clients/data_info.clj
@@ -4,6 +4,7 @@
             [cheshire.core :as cheshire]
             [clj-http.client :as http]
             [clojure.string :as string]
+            [medley.core :refer [remove-vals]]
             [apps.util.config :as config]
             [apps.util.service :as service]))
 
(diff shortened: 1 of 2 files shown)


src/apps/clients/data_info.clj
```diff
@@ -16,20 +17,23 @@
...
+                                                    :filter-include filter-include
+                                                    :filter-exclude filter-exclude))
+                 :body         (cheshire/encode {:paths paths :ids ids})
                  :content-type :json
                  :as           :json}))))
 
 (defn get-data-ids
   [user paths]
-  (->> (get-file-stats user paths)
+  (->> (get-path-info user :paths paths :filter-include "id")
```
ianmcorvidae: It maps down to just this a few lines below, so clearly it only needs this.

src/apps/service/apps/jobs/submissions.clj
```diff
@@ -141,7 +141,7 @@
   [user submission]
   (let [output-dir (ft/build-result-folder-path submission)]
     (try+
-     (data-info/get-file-stats user [output-dir])
+     (data-info/get-path-info user :paths [output-dir] :filter-include "path")
```
ianmcorvidae: Technically I'm not sure this needs *anything*, because it doesn't use the return value unless it errors. I've put in `path` though just to have it do something (which is to say, only validate, because it doesn't need to fetch anything to get the path).

src/apps/service/apps/jobs/submissions.clj
```diff
@@ -45,7 +45,7 @@
   [user input-paths-by-id]
   (->> (flatten (vals input-paths-by-id))
        (remove string/blank?)
-       (get-file-stats user)
+       (#(get-file-stats user % :filter-include "infoType,path,file-size"))
```
ianmcorvidae: This one's the complex bit of filtering. `infoType` is used just below. `path` and `file-size` are both used by `validate-path-list-stats`. After validating, it's passed to `extract-ht-param-ids` which seems to only use the path (via `(map :path path-list-stats)`). Finally it's used with another `(map :path ` in submit-batch-job itself.

src/apps/service/apps/jobs/submissions.clj
psarando: It looks like this function only returns the `output-dir` path it's checking with `stat-gatherer`. This is probably some old code, because it seems like it could be refactored to use `existence-checker` instead (but that's not necessary for this PR).

src/apps/service/apps/jobs/submissions.clj
psarando: These filters look correct to me.
Since this is the only spot where this private `get-file-stats` function is used, the `filter-include` list could be pushed into the private `get-file-stats` function, then this line wouldn't have to change.

src/apps/service/apps/jobs/submissions.clj
ianmcorvidae: Ah, yeah, I think you're right. This'll do for now.

src/apps/service/apps/jobs/submissions.clj
ianmcorvidae: That's a good idea. I'll push it up into that.

PR #74, opened 2017-07-26
CORE-8961: return a reason if an app can't be published

dennisroberts71:


dennisroberts71:
Thanks for the review!

dennisroberts71:
Thanks for the review!

PR #75, opened 2017-07-31
CORE-8936: added `GET /admin/analyses/by-external-id/:external-id` an…

dennisroberts71:
…d `POST /admin/analyses/by-external-id`


dennisroberts71:
Thanks for the review!

PR #76, opened 2017-10-03
De 1129 Webhooks endpoints

sri2k1us:
This includes 2 endpoints:
- `GET` Lists all webhooks + topic subscriptions for the user
- `PUT` Clear existing webhooks and insert new webhooks for the user


src/apps/routes/webhooks.clj
```diff
@@ -0,0 +1,37 @@
+(ns apps.routes.webhooks
+  (:use [common-swagger-api.schema]
+        [apps.routes.params]
+        [apps.routes.schemas.webhooks]
+        [apps.webhooks]
+        [ring.util.http-response :only [ok]]
+        [korma.db :only [transaction]])
+  (:import (org.postgresql.util GT)))
```
psarando: This `GT` appears to be an unused import?

src/apps/routes/webhooks.clj
```diff
@@ -0,0 +1,37 @@
...
+                :summary "List Webhooks"
+                :description "Returns all of the webhooks defined for the user."
+                (ok (list-webhooks user)))
+           (PUT "/" []
+                 :query [{:keys [user]} SecuredQueryParams]
+                 :body [body WebhookList]
+                 :return WebhookList
+                 :summary "Add a Webhook"
+                 :description "Adds a new webhook to the system."
+                 (transaction (ok (add-webhooks user body))))
```
psarando: I think it would be cleaner to keep this `transaction` inside the `add-webhooks` service function.

src/apps/routes/webhooks.clj
```diff
@@ -0,0 +1,37 @@
...
+           (context "/topics" []
+                    (GET "/" []
+                         :query [params SecuredQueryParams]
+                         :return TopicList
+                         :summary "List notification topics"
+                         :description "Returns all of the notification topics defined"
+                         (ok (list-topics)))
+                    ))
```
psarando: Formatting mishap 💥

src/apps/webhooks.clj
```diff
@@ -0,0 +1,61 @@
...
+    (when topic_id
+      (insert :webhooks_subscription (values {:webhook_id (log/spy webhook_id)
+                                              :topic_id   topic_id})))))
+(defn add-webhook [user url type topics]
+  (let [user_id (get-user-id (str user "@" (config/uid-domain)))
+        type_id (get-type-id type)
+        id (:id (insert :webhooks (values {:user_id user_id
+                                           :url     url
+                                           :type_id type_id})))]
+    (println "type_id: " type_id)
```
psarando: Looks like debugging code that could be removed.

src/apps/routes/schemas/webhooks.clj
```diff
@@ -0,0 +1,20 @@
...
+
+(def WebhookIdParam (describe UUID "A UUID that is used to identify the Webhook"))
+
+(defschema Webhook
+           {(optional-key :id) WebhookIdParam
+            :type (describe NonBlankString "Type of webhook subscription")
+            :url (describe NonBlankString "Url to post the notification")
+            :topics (describe [NonBlankString] "A List of topic names")})
+(defschema WebhookList
+           {:webhooks (describe [Webhook] "A List of webhooks")})
```
psarando: The code formatting of this file is inconsistent, so it should probably be cleaned up to match other `schemas` namespaces.

src/apps/webhooks.clj
```diff
@@ -0,0 +1,61 @@
...
+    type_id
+    (cxu/bad-request (str "Unable to find webhook type: " type))))
+
+(defn- add-topic-subscription [webhook_id topic_name]
+  (let [topic_id (:id (first (select :webhooks_topic (fields :id) (where {:topic topic_name}))))]
+    (when topic_id
+      (insert :webhooks_subscription (values {:webhook_id (log/spy webhook_id)
+                                              :topic_id   topic_id})))))
+(defn add-webhook [user url type topics]
+  (let [user_id (get-user-id (str user "@" (config/uid-domain)))
```
psarando: Since `add-webhook` is only being called by `add-webhooks`, and `add-webhooks` is already looking up the `user_id`, then it could be passed directly to this function instead of the username. That way `get-user-id` only has to be called once, and not for every webhook to be added.

src/apps/webhooks.clj
```diff
@@ -0,0 +1,61 @@
...
+
+(defn get-type-id [type]
+  (if-let [type_id (:id (first (select :webhooks_type (fields :id) (where {:type type}))))]
+    type_id
+    (cxu/bad-request (str "Unable to find webhook type: " type))))
+
+(defn- add-topic-subscription [webhook_id topic_name]
+  (let [topic_id (:id (first (select :webhooks_topic (fields :id) (where {:topic topic_name}))))]
+    (when topic_id
+      (insert :webhooks_subscription (values {:webhook_id (log/spy webhook_id)
```
psarando: Looks like a `log/spy` debugging call was left behind.

src/apps/routes/webhooks.clj
sri2k1us: :man_facepalming:

src/apps/routes/webhooks.clj
sri2k1us: function `add-webhooks` already does queries under `transaction`. Can I just go ahead and remove it from here ?

psarando: Yes, that will work 👍

src/apps/webhooks.clj
sri2k1us: 👍

src/apps/routes/webhooks.clj
```diff
@@ -0,0 +1,36 @@
+(ns apps.routes.webhooks
+  (:use [common-swagger-api.schema]
+        [apps.routes.params]
+        [apps.routes.schemas.webhooks]
+        [apps.webhooks]
+        [ring.util.http-response :only [ok]]
+        [korma.db :only [transaction]]))
```
psarando: Now that the `transaction` has been removed from the route, it can be removed from this `:use`.

sri2k1us: 👍

PR #77, opened 2017-10-11
CORE-9016 Make HT Path List job submissions async

psarando:
This PR will migrate batch job submissions into a separate thread, allowing the `POST /analyses` endpoint to submit the parent job and respond to the client right away, while sub-jobs are submitted in the background thread.

Once the background thread submits all sub-jobs, it will send a notification to the user which includes the count of successful sub-job submissions. If an exception is caught, or if all sub-jobs fail to submit, then the parent job will be marked as failed, and the user will receive a job failure notification for the parent.

This will allow us to lift the 128-paths limit on jobs submitted with HT Path List files to something much larger. We should probably keep some limit in place for now, until we have async job submissions that are 0-downtime compatible, since these sub-job submissions will be lost and the user will not be notified if the apps service goes down during HT Path List processing.

A side-effect of these updates means the `POST /analyses` endpoint no longer always returns the `Submitted` status for every job, and may now return a `Failed` status.

### DIFF
--- a/src/apps/service/apps/jobs.clj
+++ b/src/apps/service/apps/jobs.clj
@@ -1,12 +1,8 @@
 (ns apps.service.apps.jobs
-  (:use [korma.db :only [transaction]]
-        [slingshot.slingshot :only [try+]])
+  (:use [slingshot.slingshot :only [try+]])
   (:require [clojure.tools.logging :as log]
-            [clojure-commons.error-codes :as ce]
-            [clojure-commons.exception-util :as cxu]
             [kameleon.db :as db]
             [apps.clients.notifications :as cn]
-            [apps.clients.permissions :as perms-client]
             [apps.persistence.jobs :as jp]
             [apps.service.apps.job-listings :as listings]
             [apps.service.apps.jobs.params :as job-params]
@@ -15,10 +11,6 @@
             [apps.service.apps.jobs.submissions :as submissions]
             [apps.service.apps.jobs.util :as ju]
             [apps.util.service :as service]))
-
-(defn supports-job-type
-  [apps-client job-type]
-  (contains? (set (.getJobTypes apps-client)) job-type))
 
 (defn get-unique-job-step
   "Gets a unique job step for an external ID. An exception is thrown if no job step
@@ -181,10 +173,7 @@
 
 (defn submit
   [apps-client user submission]
-  (transaction
-   (let [job-info (submissions/submit apps-client user submission)]
-     (perms-client/register-private-analysis (:shortUsername user) (:id job-info))
-     job-info)))
+  (submissions/submit apps-client user submission))
 
 (defn list-job-permissions
   [apps-client user job-ids]
--- a/src/apps/service/apps/jobs/submissions.clj
+++ b/src/apps/service/apps/jobs/submissions.clj
@@ -230,11 +230,10 @@
         {:status jp/failed-status}))))
 
 (defn- async-submit-batch-jobs
-  [apps-client user path-list-stats {job-id :parent_id :as submission}]
+  [apps-client user ht-paths {job-id :parent_id :as submission}]
   (future
     (try+
-      (let [ht-paths      (set (map :path path-list-stats))
-            path-lists    (get-path-list-contents-map user ht-paths)
+      (let [path-lists    (get-path-list-contents-map user ht-paths)
             paths-exist   (validate-path-lists user path-lists)
             path-maps     (map-slices path-lists)
             job-stats     (->> path-maps
@@ -273,10 +272,11 @@
   [apps-client user input-params-by-id input-paths-by-id path-list-stats job-types app submission]
   (pre-submit-batch-validation input-params-by-id input-paths-by-id path-list-stats)
 
-  (let [output-dir (get-batch-output-dir user submission)
+  (let [ht-paths   (set (map :path path-list-stats))
+        output-dir (get-batch-output-dir user submission)
         batch-id   (save-batch user job-types app submission output-dir)
         submission (preprocess-batch-submission submission output-dir batch-id)]
-    (async-submit-batch-jobs apps-client user path-list-stats submission)
+    (async-submit-batch-jobs apps-client user ht-paths submission)
     (job-listings/list-job apps-client batch-id)))
 
 (defn submit
--- a/src/apps/service/apps/jobs/submissions.clj
+++ b/src/apps/service/apps/jobs/submissions.clj
@@ -233,22 +233,24 @@
   [apps-client user path-list-stats {job-id :parent_id :as submission}]
   (future
     (try+
-      (let [ht-paths    (set (map :path path-list-stats))
-            path-lists  (get-path-list-contents-map user ht-paths)
-            paths-exist (validate-path-lists user path-lists)
-            path-maps   (map-slices path-lists)
-            job-stats   (doall (map-indexed (partial submit-job-in-batch apps-client user submission paths-exist)
-                                            path-maps))]
-        (when-not (some (partial = jp/submitted-status)
-                        (map :status job-stats))
+      (let [ht-paths      (set (map :path path-list-stats))
+            path-lists    (get-path-list-contents-map user ht-paths)
+            paths-exist   (validate-path-lists user path-lists)
+            path-maps     (map-slices path-lists)
+            job-stats     (->> path-maps
+                               (map-indexed (partial submit-job-in-batch apps-client user submission paths-exist))
+                               doall)
+            success-count (->> job-stats
+                               (filter (comp (partial = jp/submitted-status) :status))
+                               count)]
+        (if (> success-count 0)
+          (notifications/send-batch-submission-completed user (job-listings/list-job apps-client job-id) success-count)
           (throw+ "all batch sub-job submissions failed.")))
       (catch Object _
         (log/error (:throwable &throw-context)
                    "batch job submission failed.")
         (jp/update-job job-id jp/failed-status nil)
-        (notifications/send-job-status-update (:shortUsername user)
-                                              (:email user)
-                                              (job-listings/list-job apps-client job-id))))))
+        (notifications/send-job-status-update user (job-listings/list-job apps-client job-id))))))
 
 (defn- preprocess-batch-submission
   [submission output-dir parent-id]
--- a/src/apps/service/apps/jobs/submissions.clj
+++ b/src/apps/service/apps/jobs/submissions.clj
@@ -14,6 +14,8 @@
             [apps.persistence.app-metadata :as ap]
             [apps.persistence.jobs :as jp]
             [apps.service.apps.job-listings :as job-listings]
+            [apps.service.apps.jobs.submissions.async :as async]
+            [apps.service.apps.jobs.submissions.submit :as submit]
             [apps.util.config :as config]
             [apps.util.service :as service]))
 
@@ -24,13 +26,6 @@
        (filter (comp type-set :type))
        (map (juxt (comp keyword :id) identity))
        (into {})))
-
-(defn- submit-and-register-private-job
-  [apps-client user submission]
-  (transaction
-    (let [job-info (.submitJob apps-client submission)]
-      (perms-client/register-private-analysis (:shortUsername user) (:id job-info))
-      job-info)))
 
 (defn- get-file-stats
   "Gets information for the provided paths. Filters to only path, infoType, and file size."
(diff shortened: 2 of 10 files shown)


src/apps/service/apps/jobs.clj
```diff
@@ -181,10 +173,7 @@
 
 (defn submit
   [apps-client user submission]
-  (transaction
```
psarando: This transaction had to be removed from here, since async sub-job submissions cannot share the same transaction as the parent call that responds to the client immediately, because the parent's db connection will close after the response is sent to the client.

src/apps/service/apps/jobs/submissions.clj
```diff
@@ -179,65 +117,31 @@
...
-    (dorun (map-indexed (partial submit-job-in-batch apps-client user submission paths-exist) path-maps))
-    (-> (job-listings/list-job apps-client batch-id)
-        (assoc :missing-paths (extract-missing-paths paths-exist))
-        remove-nil-values)))
+  (pre-submit-batch-validation input-params-by-id input-paths-by-id path-list-stats)
+
+  (let [ht-paths   (set (map :path path-list-stats))
+        output-dir (get-batch-output-dir user submission)
+        batch-id   (save-batch user job-types app submission output-dir)]
+    (async/submit-batch-jobs apps-client user ht-paths submission output-dir batch-id)
```
psarando: This is the call that submits the HT Path List and sub-job processing to a `future` call, but we ignore the result of that `future` so we don't block and this function can respond to the client right away.

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -0,0 +1,118 @@
...
+      {:status jp/failed-status})))
+
+(defn- preprocess-batch-submission
+  [submission output-dir parent-id]
+  (assoc submission
+    :output_dir           output-dir
+    :parent_id            parent-id
+    :create_output_subdir false))
+
+(defn submit-batch-jobs
```
psarando: All of the functions in this namespace were imported from `apps.service.apps.jobs.submissions`, but this function is slightly different than the old batch submission function.

This function wraps sub-job submissions in a `future` call, but it also catches all exceptions so it can mark the parent job as `Failed`, and it will also send notifications to the user on success or failure.

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -0,0 +1,118 @@
...
+  [max-paths first-list-path first-list-count]
+  (throw+
+    {:type       :clojure-commons.exception/illegal-argument
+     :error      (str "The HT Analysis Path List exceeds the maximum of "
+                      max-paths
+                      " allowed paths.")
+     :path       first-list-path
+     :path-count first-list-count}))
+
+(defn- validate-path-lists
```
psarando: The old `apps.service.apps.jobs.submissions/validate-path-lists` function used to also check the existence of every path in the HT Path List files, but that's been removed from this namespace since that existence checker endpoint is limited to 1000 paths, and we want to raise the HT Path List limit higher than that.

ianmcorvidae: So, just to be sure I understand, what'll happen in case a file in a pathlist doesn't exist is that the job will get submitted with the path, and then porklock will fail to download it, and the subjob will be marked failed?

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -0,0 +1,118 @@
...
+
+(defn submit-batch-jobs
+  "Asynchronously submits sub-jobs for a parent batch job,
+   where `ht-paths` is a list of paths of the HT Path List input files,
+   `submission` is the original parent submission map,
+   `output-dir` is the path to the output folder created for the parent job,
+   and `parent-id` is the parent job's ID to use for each sub-job's `parent_id` field.
+   A notification will be sent to the user on success or on failure to submit all sub-jobs."
+  [apps-client user ht-paths submission output-dir parent-id]
+  (future
```
ianmcorvidae: This is probably nitpicky, but I wonder if `future` is really the best abstraction to use here. In particular, a future is really designed for something that is going to be deref'd later -- doing computation in another thread, but not really something that autonomously does something and cleans up after itself, like this. It *works*, of course, but I think it communicates the wrong thing about the intent. Instead, I wonder if it'd be better to use either an agent or an explicit thread. For an agent, what you'd do is create an agent, then use [send-off](https://clojuredocs.org/clojure.core/send-off) or `send` to pass the function and arguments, which would then update the state of the agent. Largely, this is the same as using a future, but you could choose to not even return the agent from this function, which would make it clear the state of the agent isn't meaningful (note: futures are actually built on agents, and use the same pool as `send-off`). Using a thread directly, you'd do `(.start (Thread. (fn [] ...)))`, which doesn't even have any communication-back-to-the-parent stuff built in, so it's even more obvious. The one benefit I can see of using an agent or the future as you currently have it is that enforces the use of a thread pool, which could then be of a fixed size to limit how many huge batch jobs could be mid-submission at once, but the `send-off` pool (used by `future` as well) grows without bound from what I can read (it's using a [cached thread pool](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newCachedThreadPool()) -- if you use agents and `send`, those use a fixed thread pool). If you wanted to be really fancy, you could create a new thread pool of a fixed size or caching like the send-off one, and with a naming pattern that specifies their use (e.g. the send-off/future pool uses `"clojure-agent-send-off-pool-%d"` with an atomic value for the `%d`, so maybe you'd do `"apps-submit-batch-jobs-%d"` or something), then create agents and use `send-via` to delegate the execution to that thread pool.

But it feels to me like the best expression of intent would be `(.start (Thread. (fn [] ...)))`.

ianmcorvidae: additional note, if you like the whole named thread thing but thing doing a full pool is overkill (like I suggest), you can still pass a name as the second argument to the `Thread` constructor. Could do something like name the threads `"apps-submit-batch-jobs-<uuid of parent job>"`, then the logs will be real clear about what's going on! (and note that threads can have the same name, so don't worry about potential conflicts probably, but it's still better if they are unique in my opinion)

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -0,0 +1,118 @@
...
+            job-stats     (->> path-maps
+                               (map-indexed (partial submit-job-in-batch apps-client user submission))
+                               doall)
+            success-count (->> job-stats
+                               (filter (comp (partial = jp/submitted-status) :status))
+                               count)]
+        (if (> success-count 0)
+          (notifications/send-batch-submission-completed user
+                                                         (job-listings/list-job apps-client parent-id)
+                                                         success-count)
```
ianmcorvidae: I think it might be nice if this notification could specify not just how many succeeded, but also how many failed. Then just looking at the notification it's easy to tell if you can pretty much just trust it to run, or if you should go investigate why some of them failed. Not sure how hard that'd be to implement or if that's better done with a later implementation, but I thought I'd throw it out there.

src/apps/service/apps/jobs/submissions/async.clj
ianmcorvidae: Secondarily, since this is now done async anyway, how bad would it be to just [partition-all](https://clojuredocs.org/clojure.core/partition-all) the list of paths into sets of 1000 or fewer and still check for existence? Letting it fail once it hits condor seems like it'd at least work, but with the relieving of the time pressure by putting this in a separate thread it feels weird to also remove validation like this. But maybe pushing that check out to condor (implicitly) makes the apps service less of a bottleneck? Not sure.

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -0,0 +1,118 @@
...
+               :error "All HT Analysis Path Lists must have the same number of paths."}))
+    path-lists))
+
+(defn- get-path-list-contents
+  [user path]
+  (try+
+    (when (seq path) (data-info/get-path-list-contents user path))
+    (catch Object _
+      (log/error (:throwable &throw-context)
+                 "job submission failed: Could not get file contents of path list input.")
```
ianmcorvidae: Will the throwable contain the path which failed to be loaded, or would it make sense to put the path which failed into the message?

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -0,0 +1,118 @@
...
+   and `parent-id` is the parent job's ID to use for each sub-job's `parent_id` field.
+   A notification will be sent to the user on success or on failure to submit all sub-jobs."
+  [apps-client user ht-paths submission output-dir parent-id]
+  (future
+    (try+
+      (let [path-lists    (validate-path-lists (get-path-list-contents-map user ht-paths))
+            path-maps     (map-slices path-lists)
+            submission    (preprocess-batch-submission submission output-dir parent-id)
+            job-stats     (->> path-maps
+                               (map-indexed (partial submit-job-in-batch apps-client user submission))
```
psarando: I started working on `CORE-9014` and I realized that `submit-job-in-batch` will have to check the parent job's status before submitting each sub-job, in case the user tries to cancel the parent in the middle of batch submissions.

I'll go ahead and make that change for this PR.

src/apps/service/apps/jobs/submissions.clj
```diff
@@ -179,65 +117,31 @@
...
+  (transaction
+    (let [batch-id (save-batch* user app submission output-dir)]
+      (save-batch-step batch-id (first job-types))
+      (perms-client/register-private-analysis (:shortUsername user) batch-id)
+      batch-id)))
+
+(defn- pre-submit-batch-validation
+  [input-params-by-id input-paths-by-id path-list-stats]
+  (doseq [path-stat path-list-stats]
+    (validate-path-list-stats path-stat))
```
slr71: I wonder if it might make sense to remove the path list file size validation and rely solely on the path list length validation. If we're going to modify the list length frequently then having to modify this setting in lockstep might be a bit of a pain. The one advantage to keeping this setting is that it allows us to constrain the amount of memory consumed by a single path list somewhat.

psarando: For now I was going to set the file size limit to something large but not too large. A path list I've been using in testing has just over 30k paths and is just under 4MB, so I figure a limit of 64MB should be enough for anybody 😉  That lets us adjust the path limit without having to also adjust the size limit, but the size limit is there in case someone has very long paths.

I just realized that I could lookup the iRODS path length limit, and see how big of a file that would create with something like 100k paths. If that's reasonably small, then I agree that we can get rid of one limit and keep the other.

src/apps/service/apps/jobs/submissions/async.clj
psarando: Yeah, I'm not sure either. Before we would not submit jobs for missing paths, but I don't think the user ever got a message about the missing paths (the service returned the list of missing paths, but I don't think that was displayed in the UI). So if 1 out of 3 paths was missing, their batch job would successfully complete, but it would only have 2 sub-jobs. I'm not sure that's very UX friendly. At least with the failed job, the user could examine the error logs and see why the job failed (file not found).

If we can find a user-friendly way of letting the user know which paths are missing, then I think we can add back existence checking (whether in batches of 1000 or once per sub-job).

src/apps/service/apps/jobs/submissions/async.clj
psarando: I think I was mostly swayed by this SO answer and went with `future` for simplicity: https://stackoverflow.com/a/8291914/1884059

I see now what you mean by the name of the thread in the logs, and I like the idea of tracking async submissions in the logs, so I think I will go ahead and switch to using `Thread` named with the parent job's ID and adding some info logging 👍

src/apps/service/apps/jobs/submissions/async.clj
ianmcorvidae: That makes sense. I suppose this may just be a better thing to do later; letting the user know which paths are missing seems like it could go in notifications, but it could get too noisy. Maybe the best eventual place for this would be if we eventually add something like a job history view, we could send one notification if any paths are missing, which sends the user to the history (which should have events marked for the missing files).

Anyway, now that you've reminded me of the prior behavior of just not submitting jobs, I do think you're right and this is better.

src/apps/service/apps/jobs/submissions.clj
psarando: After some discussion in chat with @ianmcorvidae, we decided we could get rid of the max HT Path file size limit as a config setting, and instead calculate a reasonable limit based on the max paths config setting. This helps prevent loading a massive, non-HT-Path-List file into memory by mistake (since users can set any file they wish as the HT Path List file type). In fact, it would also prevent loading a massive HT Path List file that goes way beyond the paths limit, and we wouldn't have to count the paths first before giving the client an error response.

ianmcorvidae:
Changes since I last looked look good! :shipit:

psarando:
Thanks for the reviews! 👍

PR #78, opened 2017-10-19
Allow the `GET /admin/apps` endpoint to list private apps as well as public apps

slr71:


slr71:
Thanks for the review!

PR #79, opened 2017-10-20
CORE-9014 Update analyses stop endpoint to also stop all child jobs.

psarando:
This PR will update `apps.service.apps.jobs/stop-job` to also stop all child jobs, even if the parent job has already been stopped in a previous request.

Stopping child jobs will run async in a named thread, allowing the endpoint to respond to the client right away.

### DIFF
--- a/src/apps/service/apps/jobs.clj
+++ b/src/apps/service/apps/jobs.clj
@@ -1,6 +1,8 @@
 (ns apps.service.apps.jobs
-  (:use [slingshot.slingshot :only [try+]])
-  (:require [clojure.tools.logging :as log]
+  (:use [korma.db :only [transaction]]
+        [slingshot.slingshot :only [try+]])
+  (:require [clojure.string :as string]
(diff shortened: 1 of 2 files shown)


src/apps/service/apps/jobs.clj
```diff
@@ -32,16 +34,19 @@
   [job-id]
   (service/assert-found (jp/lock-job job-id) "job" job-id))
 
+(defn- format-job-for-status-update
+  [apps-client job-id]
+  (let [job        (jp/get-job-by-id job-id)
+        app-tables (.loadAppTables apps-client [job])
+        rep-steps  (jp/list-representative-job-steps [job-id])
+        rep-steps  (group-by (some-fn :parent_id :job_id) rep-steps)]
+    (listings/format-job apps-client nil app-tables rep-steps job)))
```
slr71: Good refactor. 👍

psarando:
Thanks for the review! 👍

PR #80, opened 2017-10-20
CORE-9046: allow apps that already have documentation to be published

slr71:


src/apps/service/apps/de/docs.clj
```diff
@@ -58,3 +58,16 @@
...
+  [{:keys [username]} app-id {docs :documentation}]
+  (if-let [current-docs (dp/get-documentation app-id)]
+    (dp/edit-documentation (v/get-valid-user-id username) docs app-id)
+    (dp/add-documentation (v/get-valid-user-id username) docs app-id)))
+
+(defn has-docs?
+  "Determines whether or not an app has docs already."
+  [app-id]
+  (when (dp/get-documentation app-id)
+    true))
```
psarando: I might have just gone with `(def has-docs? dp/get-documentation)` :man_shrugging:

src/apps/routes/schemas/app.clj
```diff
@@ -549,15 +549,15 @@
...
 
 (defschema PublishAppRequest
   (-> AppBase
     (->optional-param :id)
     (->optional-param :name)
     (->optional-param :description)
-    (assoc :documentation AppDocParam
-           :references AppReferencesParam)
+    (assoc (optional-key :documentation) AppDocParam
+           (optional-key :references) AppReferencesParam)
```
psarando: If the `:avus` and `:references` keys are also becoming optional, then I think we need to validate them in the request similar to these changes for `:documentation` (i.e. the request should only be valid if those items are included or are already set in the db).

Otherwise, maybe we should keep the categories and references required?

src/apps/service/apps/de/docs.clj
slr71: Yes, that would work just as well.

src/apps/routes/schemas/app.clj
slr71: The behavior of the endpoint really hasn't changed significantly, though. There was nothing preventing someone from calling the endpoint with empty lists for both `:avus` and `:references` before. This just makes it easier for someone building a request to do so without adding these fields. I could add the validation, but doing so is a bigger departure than the code that is already there.

psarando: Oops, you're right, I overlooked the fact that these were array parameters.
Let's go without the validation then 👍

slr71:
Thanks for the review!

PR #81, opened 2017-10-25
CORE-9052: add the ability to search for app categories by name

slr71:
This is a new endpoint to search for app categories by name.

### DIFF
--- a/src/apps/persistence/app_groups.clj
+++ b/src/apps/persistence/app_groups.clj
@@ -34,6 +34,16 @@
                  (fields :id :name :is_public)
                  (where {:id app_group_id}))))
 
+(defn search-app-groups
+  "Searches for app category information by name."
+  [names]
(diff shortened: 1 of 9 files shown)


src/apps/persistence/app_groups.clj
```diff
@@ -34,6 +34,16 @@
                  (fields :id :name :is_public)
                  (where {:id app_group_id}))))
 
+(defn search-app-groups
+  "Searches for app category information by name."
+  [names]
+  (-> (select* [:app_categories :c])
```
psarando: Keep in mind that public apps these days use ontology metadata for categories, and not these old db categories.

Is this for QA tests?

slr71: Yeah, this is for QA tests.

slr71:
Thanks for the review.

PR #83, opened 2017-10-31
CORE-9059: added an option to permanently delete a reference genome

slr71:


slr71:
Thanks for the review!

PR #84, opened 2017-10-31
de-1129 Add endpoint to list webhook types.

sri2k1us:


src/apps/webhooks.clj
```diff
@@ -58,3 +63,6 @@
 
 (defn list-topics []
   {:topics (select [:webhooks_topic :wt])})
+
+(defn list-types []
+  {:webhooktypes (select [:webhooks_type :wtypes])})
```
slr71: There's no need to change this, but a table alias isn't really needed in cases where you're not doing a join.

src/apps/webhooks.clj
```diff
@@ -13,8 +13,13 @@
                       (fields [:wt.topic])
                       (where {:ws.webhook_id webhook-id}))))
 
-(defn- format-webhook [{:keys [id] :as webhook}]
-  (assoc webhook :topics (get-webhook-topics id)))
+(defn- get-webhook-type [type]
+  (first (select [:webhooks_type :wt] (where {:wt.type type}))))
```
slr71: This doesn't have to be changed, but a table alias isn't really necessary when you're not doing any joins. For example, you could change this function to something like this:

```clojure
(defn- get-webhook-type [type]
  (first (select :webhooks_type (where {:type type}))))
```

The alias doesn't do any harm, of course. It's just that leaving it off saves a little bit of typing. 😄

src/apps/routes/webhooks.clj
```diff
@@ -26,6 +26,14 @@
...
                          (ok (list-topics)))
+                    )
+           (context "/types" []
+                    (GET "/" []
+                         :query [params SecuredQueryParams]
+                         :return WebhookTypeList
+                         :summary "List webhooks types"
+                         :description "Returns all webhook types defined"
+                         (ok (list-types)))
                     ))
```
slr71: This is so minor that I feel bad mentioning it, but the convention for closing parentheses in Clojure is to put them at the end of the form rather than to put them on a line by themselves.

sri2k1us: 👍

src/apps/webhooks.clj
sri2k1us: 👍

src/apps/webhooks.clj
sri2k1us: 👍

sri2k1us:
They didn't cause any errors. I changed it anyway to make the code little more terse.

slr71:
Ah, good to know. I could have sworn that I remembered getting a null pointer exception when I did that once. Maybe it was a previous version of Korma...or maybe I was thinking of something else entirely. 😆

slr71:
The change looks good to me. Merge it whenever you're ready.

src/apps/routes/schemas/webhooks.clj
```diff
@@ -3,11 +3,15 @@
         [schema.core :only [defschema optional-key]])
   (:import [java.util UUID]))
 
-(def WebhookIdParam (describe UUID "A UUID that is used to identify the Webhook"))
+(defschema WebhookType
+  {:id       (describe UUID "A UUID for the type")
+   :type     (describe NonBlankString "Webhook type")
+   :template (describe NonBlankString "Template for this Webhook type")})
```
psarando: This makes the `template` required in both the webhook list responses and requests.

Are we allowing the client to update the `template` for webhook types?
If not, we can make this field optional here so that it's not required in requests.

See also my question in cyverse-de/ui#242

PR #85, opened 2017-11-01
CORE-9055: admin workspace listing and deletion endpoints.

slr71:


slr71:
Thanks for the review!

PR #86, opened 2017-11-02
CORE-9057 Add Multi-Input Path List job submission support

psarando:
This PR will add job submission support for a new `multi-input-path-list` file info-type, which is only valid for use in Multi-Input app parameters.

This info-type is defined in cyverse-de/heuristomancer#4.

The current strategy for `Multi-Input Path List` job submission support is to have this service "expand" the path-list contents in the input parameter job submission JSON, by replacing the original path with the downloaded list of paths.

This is similar to the strategy used for HT Path List files, except the "Multi-Input Path List expansion" does not need to be asynchronous, since the replacement can be done in a reasonable amount of time.

A bigger consideration is that a path list file of 100k paths can have a file size around 100MB, which means the JSON submission to the job services will also be at least this big, which may be too much for our job submission queues. So I will leave this PR open until we can decide if an alternative can be implemented fairly quickly, otherwise this will be the temporary solution (which should have a limit of about 20k paths) until the async job submission service is implemented.

### DIFF
--- a/src/apps/persistence/jobs.clj
+++ b/src/apps/persistence/jobs.clj
@@ -207,6 +207,12 @@
       first
       :status))
 
+(defn- submission->save-format
+  [submission]
+  (-> submission
+      (dissoc :job_config)
+      cheshire/encode))
+
 (defn save-job
   "Saves information about a job in the database."
   [{system-id :system_id username :username :as job-info} submission]
@@ -214,7 +220,7 @@
       (assoc :job_type_id (job-type-id-from-system-id system-id)
              :user_id     (get-user-id username))
       remove-nil-values
-      (save-job-with-submission (cheshire/encode submission))))
+      (save-job-with-submission (submission->save-format submission))))
 
 (defn save-job-step
   "Saves a single job step in the database."
--- a/src/apps/persistence/jobs.clj
+++ b/src/apps/persistence/jobs.clj
@@ -207,12 +207,6 @@
       first
       :status))
 
-(defn- submission->save-format
-  [submission]
-  (-> submission
-      (dissoc :job_config)
-      cheshire/encode))
-
 (defn save-job
   "Saves information about a job in the database."
   [{system-id :system_id username :username :as job-info} submission]
@@ -220,7 +214,7 @@
       (assoc :job_type_id (job-type-id-from-system-id system-id)
              :user_id     (get-user-id username))
       remove-nil-values
-      (save-job-with-submission (submission->save-format submission))))
+      (save-job-with-submission (cheshire/encode submission))))
 
 (defn save-job-step
   "Saves a single job step in the database."
--- a/src/apps/service/apps/agave/jobs.clj
+++ b/src/apps/service/apps/agave/jobs.clj
@@ -21,13 +21,15 @@
          :query "status=${JOB_STATUS}&external-id=${JOB_ID}&end-time=${JOB_END_TIME}")))
 
 (defn- format-submission
-  [agave job-id result-folder-path submission]
-  (->> (assoc (dissoc submission :starting_step :step_number)
-         :callbackUrl          (build-callback-url job-id)
-         :job_id               job-id
-         :step_number          (:step_number submission 1)
-         :output_dir           result-folder-path
-         :create_output_subdir false)))
+  [agave job-id result-folder-path {:keys [config] :as submission}]
+  (-> submission
+      (dissoc :starting_step :step_number :job_config)
+      (assoc :config               (:job_config submission config)
+             :callbackUrl          (build-callback-url job-id)
+             :job_id               job-id
+             :step_number          (:step_number submission 1)
+             :output_dir           result-folder-path
+             :create_output_subdir false)))
 
 (defn- prepare-submission
   [agave job-id submission]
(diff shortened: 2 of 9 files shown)


src/apps/service/apps/jobs/submissions.clj
```diff
@@ -147,12 +178,60 @@
...
+    (config->expand-multi-input-path-lists config
+                                           user
+                                           input-params-by-id
+                                           input-paths-by-id
+                                           multi-input-path-list-stats)
+    config))
+
+(defn- pre-process-submission
+  [{:keys [config] :as submission} user input-params-by-id input-paths-by-id path-stats]
+  (assoc submission :job_config (process-job-config config user input-params-by-id input-paths-by-id path-stats)))
```
psarando: The Multi-Input Path List input parameters are expanded in a new `job_config` key of the submission.
This allows the user's original submission config to be saved to the db without all the expanded paths,
while the `job_config` key will replace the `config` in the submission that is serialized for the job services.

src/apps/persistence/jobs.clj
```diff
@@ -207,14 +207,20 @@
...
+
 (defn save-job
   "Saves information about a job in the database."
   [{system-id :system_id username :username :as job-info} submission]
   (-> (select-keys job-info job-fields)
       (assoc :job_type_id (job-type-id-from-system-id system-id)
              :user_id     (get-user-id username))
       remove-nil-values
-      (save-job-with-submission (cheshire/encode submission))))
+      (save-job-with-submission (submission->save-format submission))))
```
psarando: This new format function will remove the new `job_config` key from the submission before saving to the db.

src/apps/service/apps/de/jobs.clj
```diff
@@ -113,8 +113,9 @@
     (submit-standalone-job user submission job)))
 
 (defn- prep-submission
-  [submission]
+  [{:keys [config] :as submission}]
   (assoc submission
+    :config               (:job_config submission config)
```
psarando: Here is where the new `job_config` key replaces the user's original `config` for DE job submission JSON.

src/apps/service/apps/combined/jobs.clj
```diff
@@ -74,9 +74,10 @@
        (validate-job-steps app-id)
        (map (partial build-job-step-save-info job-id))))
 
-(defn- prepare-de-job-step-submission
-  [job-info job-step submission]
+(defn- prepare-job-step-submission-common
+  [job-info job-step {:keys [config] :as submission}]
   (assoc submission
+    :config               (:job_config submission config)
```
psarando: Here is where the new `job_config` key replaces the user's original `config` for combined-pipeline job submission JSON.

src/apps/service/apps/agave/jobs.clj
```diff
@@ -21,13 +21,15 @@
...
-  (->> (assoc (dissoc submission :starting_step :step_number)
-         :callbackUrl          (build-callback-url job-id)
-         :job_id               job-id
-         :step_number          (:step_number submission 1)
-         :output_dir           result-folder-path
-         :create_output_subdir false)))
+  [agave job-id result-folder-path {:keys [config] :as submission}]
+  (-> submission
+      (dissoc :starting_step :step_number :job_config)
+      (assoc :config               (:job_config submission config)
```
psarando: Agave apps don't currently have multi-input parameters, but if they ever get added, then this is where the new `job_config` key will replace the user's original `config` for Agave job submission JSON.

ianmcorvidae:
Hm, in the category of "dumb, off the top of the head ideas, but which might fix issues": what if, like HT pathlists, this system created a parent and a child job, but in this case only the one child. Then, the parent job would show the pathlist as its input, and the child job would show the actual files used. Then, there's a record in our database of everything used for the job (as there is with HT pathlists), but we can still have the relaunch (of the parent job) behave like I think is desired. Or, is there more reason than that for avoiding listing out the actual files used?

psarando:
The initial use-case we received that prompted this feature was the need for ~15k input files to be combined by a single job. I don't think it's possible to display that many input files from the re-launch info in the multi-input selector without hanging the browser for a long time (spinning pinwheel of death).

The apps service will log the final job submission JSON before submitting it to the job services (but those logs probably don't last). Maybe we could store the `job_config` in the database anyway, as a record of the actual inputs used, but we don't have to send that to the client from the `relaunch` endpoint.

src/apps/service/apps/combined/jobs.clj
```diff
@@ -74,28 +74,28 @@
        (validate-job-steps app-id)
        (map (partial build-job-step-save-info job-id))))
 
-(defn- prepare-de-job-step-submission
-  [job-info job-step submission]
+(defn- prepare-common-job-step-submission
+  [job-info job-step {:keys [config] :as submission}]
   (assoc submission
+    :config               (:job_config submission config)
```
psarando: Here is where the new `job_config` key replaces the user's original `config` for combined-pipeline job submission JSON.

psarando:
@ianmcorvidae you are correct that with HT Path Lists we create separate sub-jobs for each path in the list, so we do have a record of the actual jobs run with that type of list.

I just added a commit to also save the post-processed `job_config` with the submission JSON in the database, so now we also have a record of the actual paths used for these new Multi-Input Path List files.

It also turns out I didn't have to modify the `relaunch-info` endpoint, since it doesn't return everything in the submission anyway (it already specifically returns the user's original `config` in responses).

Thanks for the reviews! 👍

PR #88, opened 2017-11-21
CORE-9073: added an optional parameter to the permission-lister endpo…

slr71:
…ints to allow them to optionally include the permissions of the authenticated user


slr71:
Thanks for the review!

slr71:
Thanks for the review!

PR #89, opened 2017-11-29
CORE-9082 Prevent PSQLException in the GET /analyses endpoint when user has 32k+ jobs/sub-jobs

psarando:
This PR will add an `apps.persistence.util/sqlfn-any-array` for use in a `where` clause, in order to prevent a `PSQLException` when passing too many items in a `(when {:id [in id-list]})` clause.

The `PSQLException` (caused by `java.io.IOException`: "Tried to send an out-of-range integer as a 2-byte value") is thrown because the JDBC driver is hitting a limit on the number of bind variables per PostgreSQL statement.

### DIFF
--- a/src/apps/persistence/util.clj
+++ b/src/apps/persistence/util.clj
@@ -1,0 +1,25 @@
+(ns apps.persistence.util
+  (:use [korma.db :only [transaction *current-conn*]])
+  (:require [korma.core :as sql]))
+
+(defn sql-array
+  "Returns a SQL ARRAY(...) object,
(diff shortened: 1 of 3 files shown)


src/apps/persistence/util.clj
```diff
@@ -0,0 +1,24 @@
+(ns apps.persistence.util
+  (:use [korma.db :only [transaction *current-conn*]])
+  (:require [korma.core :as sql]))
+
+(defn sql-array
```
slr71: 👍

psarando:
Thanks for the reviews! 👍

PR #90, opened 2017-12-01
CORE-9084 Speed up `GET /analyses` listing endpoint.

psarando:
This PR will update the `GET /analyses` listing endpoint to use SQL queries to count the statuses for child jobs for batch job listings.

This will speed up the `GET /analyses` endpoint when listing large batch jobs.


slr71:
Looks good to me. :shipit:

psarando:
Thanks for the review! 👍

PR #91, opened 2017-12-07
CORE-9086 Fix `job_config` in `format-submission-in-batch` function.

psarando:
This PR will fix the `POST /analyses` endpoint for HT Path List file inputs.

A bug was introduced to batch job processing by #86, which required a new `job_config` submission setting for Multi-Input Path List inputs.

Since #86, the user's original submission `config` is stored in the database for re-launching jobs, but the new `job_config` submission setting is used to submit actual parameters to the job services. The `job_config` submission setting is just a copy of the original submission `config` for jobs without any Multi-Input Path List inputs.

This PR will fix the `apps.service.apps.jobs.submissions.async/format-submission-in-batch` function so that it also substitutes HT path inputs in each sub-job's `job_config` submission settings.

### DIFF
--- a/src/apps/service/apps/jobs/submissions/async.clj
+++ b/src/apps/service/apps/jobs/submissions/async.clj
@@ -51,12 +51,15 @@
        (into {})))
 
 (defn- format-submission-in-batch
-  [submission job-number path-map]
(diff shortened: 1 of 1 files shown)


psarando:
Thanks for the review 👍

PR #92, opened 2018-01-18
CORE-7576 Update (un)share analyses to allow input (un)share errors.

psarando:
The `/analyses/sharing` and `/analyses/unsharing` endpoints have been updated to return error messages for any input (un)share errors, but job (un)shares can still be marked as a `success` with those errors.

Most (un)sharing validation functions have been refactored to throw exceptions for errors, which are caught in the calling (un)sharing function and formatted as a failure message. Also, these endpoints used to return only the first input (un)sharing failure message, but this PR collects all failure messages and returns them in the response.


psarando:
I'll go ahead and merge this now, but reviews are still welcome.

PR #93, opened 2018-04-02
core-8430 Add date filtering option for app stats.

sri2k1us:


src/apps/persistence/app_listing.clj
```diff
@@ -223,15 +224,21 @@
       (add-app-group-plus-public-apps-where-clause app-group-id username public-app-ids)
       (select))))
 
+(defn- add-date-limits-where-clause
+  [query {:keys [start_date end_date]}]
+  (if (and start_date end_date)
+    (where query {:end_date [between [(date->timestamp start_date) (date->timestamp end_date)]]}) query)
```
ianmcorvidae: I don't know if this is worth the bother, but it might be a bit confusing that if you don't provide both a start and end date, they'll be ignored. (May depend on what the UI/belphegor is doing with this, though). It should be equivalent to `BETWEEN` to add a `<=` and a `>=` separately, if you think it matters enough to change.

src/apps/routes/schemas/app.clj
```diff
@@ -464,7 +464,10 @@
          {(optional-key :search)
           (describe String
             "The pattern to match in an App's Name, Description, Integrator Name, or Tool Name.")
-
+          (optional-key :start_date)
+          (describe Date "The App's stat start date")
+          (optional-key :end_date)
+          (describe Date "The App's stat end date")
```
psarando: I think it might be easier to read if there are blank lines between each key-value pair, since each key-value is already listed on separate lines.

Also, it might be clearer if the docs mention that these dates will filter the app's stats. Maybe something like "Filters out the App's stats before this `start_date`" and "Filters out the App's stats after this `end_date`"?

src/apps/service/apps/combined.clj
```diff
@@ -57,7 +57,7 @@
          (util/combine-app-listings params)))
 
   (adminSearchApps [_ search-term params]
-    (->> (map #(.adminSearchApps % search-term (select-keys params [:search :app-subset])) clients)
+    (->> (map #(.adminSearchApps % search-term (select-keys params [:search :app-subset :start_date :end_date])) clients)
```
psarando: It looks like these query params will be passed to Agave app listings, so we should make sure Agave app listings will still work with these start and end date params, or make sure they are filtered out of the listing query.

Also, Agave apps with their job stats are included in this admin listing endpoint:
https://github.com/cyverse-de/apps/blob/2394f162b591f5edb4411542794c183d3cc872fc/src/apps/service/apps/agave/listings.clj#L11-L21
So do we also want to filter Agave job stats by these dates?

psarando: It looks like these extra params won't interfere with Agave app listings:
https://github.com/cyverse-de/mescal/blob/dcd101c7b6d6066c386a4de0729af6cf363663b3/src/mescal/agave_v2.clj#L97-L99

But I think we probably want to also filter Agave app stats with these start/end date params.

sri2k1us: Sounds good. I will add date filtering to agave stats and option for sending either `start_date` or `end_date`. I still don't understand why my Intellij / clojure cursive won't format code properly.

src/apps/persistence/app_listing.clj
```diff
@@ -242,17 +244,19 @@
 
 (defn- get-admin-job-stats-fields
   "Adds query fields via subselects for an app's job_count, job_count_failed, and last_used timestamp."
-  [query]
+  [query query-ops]
```
psarando: I think you meant `query-opts` (like below) here and in `get-job-stats-fields`, but that doesn't need to be changed now.

src/apps/service/apps/agave/listings.clj
```diff
@@ -29,14 +29,14 @@
 (defn get-app-details
   [agave app-id admin?]
   (-> (.getAppDetails agave app-id)
-      (add-agave-job-stats admin?)
+      (add-agave-job-stats () admin?)
```
psarando: Usually we just pass `nil` as the empty value. That's also safe for the `add-date-limits-where-clause` destructuring (`start_date` and `end_date` would just be `nil` also).

src/apps/service/apps/de/listings.clj
```diff
@@ -498,7 +498,7 @@
              :description          (:description details "")
              :references           (map :reference_text (:app_references details))
              :tools                (map format-app-tool tools)
-             :job_stats            (format-app-details-job-stats (str app-id) admin?)
+             :job_stats            (format-app-details-job-stats (str app-id) () admin?)
```
psarando: This could also be `nil` if you want to change it.

src/apps/persistence/app_listing.clj
slr71: I'll second this one. `query-ops` makes me think, `query-operations.`

src/apps/service/apps/agave/listings.clj
slr71: Ah, good point. 👍

PR #94, opened 2018-04-11
Add support for OSG singularity image paths.

slr71:


slr71:
Thanks for the review!

PR #95, opened 2018-04-11
CORE-9189 Update to create tickets for inputs and output folders for use in OSG submissions

psarando:
This PR will add `apps.service.apps.de.jobs.io-tickets/update-with-download-tickets` for use with
`apps.service.apps.de.jobs.protocol/buildInputs` and `update-with-output-dir-ticket` for use with `buildSubmission`.

These new functions will use a new `apps.clients.data-info/create-tickets` function for calling the `data-info` endpoint.

Also refactored `apps.service.apps.jobs.submissions/get-batch-output-dir` into a common `apps.service.apps.jobs.util/create-output-dir` function, for use in both `update-with-output-dir-ticket` and batch output folder creation.

### DIFF
--- a/src/apps/service/apps/de/jobs/io_tickets.clj
+++ b/src/apps/service/apps/de/jobs/io_tickets.clj
@@ -1,0 +1,42 @@
+(ns apps.service.apps.de.jobs.io-tickets
+  (:require [apps.clients.data-info :as data-info]
+            [apps.service.apps.jobs.util :as util]))
+
+(defn update-with-download-tickets
+  "Takes a list if input params,
+   creates a download ticket for the path of each input's :value,
+   and updates those inputs with the created ticket in a :ticket field.
+   For use with apps.service.apps.de.jobs.protocol/buildInputs"
+  [user params]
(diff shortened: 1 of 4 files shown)


src/apps/service/apps/de/jobs/io_tickets.clj
```diff
@@ -0,0 +1,42 @@
...
+  (if (or (empty? params) (every? (comp empty? :value) params))
+    params
+    (let [paths      (remove empty? (map :value params))
+          ticket-map (->> (data-info/create-tickets user
+                                                    paths
+                                                    :mode       "read"
+                                                    :public     false
+                                                    :uses-limit 1)
+                          :tickets
+                          (map #(vector (:path %) (:ticket-id %)))
```
slr71: I like to use `juxt` for stuff like this:

```clojure
(map (juxt :path :ticket-id))
```

psarando: Yes, that's a much better pattern. Thanks for the reminder 👍

psarando:
Thanks for the review 👍

PR #97, opened 2018-04-30
Changes related to interactive apps

johnworth:
Adds the `interactive` boolean field to the container settings map.

Adds the `interactive_apps` map to the container settings.

Adds the `container_ports` list to the container-settings.

### DIFF
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -639,16 +639,14 @@
                                :id                    (uuidify port-uuid)})))))
 
 (defn add-port
-  [settings-uuid port-map]
-  (let [host-port      (:host_port port-map)
-        container-port (:container_port port-map)
-        bind-to-host   (:bind_to_host port-map)]
-    (if (port-mapping? settings-uuid host-port container-port bind-to-host)
-      (throw (Exception. (str "port mapping already exists: " settings-uuid " " host-port " " container-port " " bind-to-host))))
-    (insert container-ports
-            (values (merge
-                     (select-keys port-map [:host_port :container_port :bind_to_host])
-                     {:container_settings_id (uuidify settings-uuid)})))))
+  [settings-uuid {host-port :host_port container-port :container_port bind-to-host :bind_to_host :as port-map}]
+  (if (port-mapping? settings-uuid host-port container-port bind-to-host)
+    (throw (Exception. (str "port mapping already exists: " settings-uuid " " host-port " " container-port " " bind-to-host))))
+  (insert container-ports
+          (values (merge
+                   (select-keys port-map [:host_port :container_port :bind_to_host])
+                   {:container_settings_id (uuidify settings-uuid)}))))
+
 
 (defn modify-port
   [settings-uuid port-uuid port-map]
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -471,7 +471,7 @@
                      (fields :host_port :container_port :bind_to_host :id))
                    (where {:tools_id id}))
            first
-           (add-interapps-info)
+           add-interapps-info
            (update :container_volumes_from add-data-container-auth :auth? auth?)
            (merge {:image (tool-image-info tool-uuid :auth? auth?)})
            filter-returns))))
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -671,14 +671,14 @@
         (dissoc (port port-uuid) :container_settings_id)))))
 
 (defn add-tool-port
-  [tool-uuid port-map]
+  [tool-uuid {host-port :host_port container-port :container_port bind-to-host :bind_to_host :as port-map}]
   (when-not (tool-has-settings? tool-uuid)
     (throw (Exception. (str "Tool " tool-uuid " does not have a container."))))
   (let [settings-uuid (tool-settings-uuid tool-uuid)]
     (dissoc
-     (if-not (port-mapping? settings-uuid (:host_port port-map) (:container_port port-map) (:bind_to_host port-map))
+     (if-not (port-mapping? settings-uuid host-port container-port bind-to-host)
        (add-port settings-uuid port-map)
-       (port-mapping settings-uuid (:host_port port-map) (:container_port port-map) (:bind_to_host port-map))))
+       (port-mapping settings-uuid host-port container-port bind-to-host)))
     :container_settings_id))
 
 (defn port-field
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -7,7 +7,8 @@
                                           container-volumes
                                           container-volumes-from
                                           container-ports
-                                          data-containers]]
+                                          data-containers
+                                          interapps-proxy-settings]]
         [apps.persistence.docker-registries :only [get-registry]]
         [apps.persistence.tools :only [update-tool]]
         [apps.util.assertions :only [assert-not-nil]]
@@ -380,6 +381,7 @@
      :name
      :entrypoint
      :tools_id
+     :interactive_apps_proxy_settings_id
      :id]))
 
 (defn add-settings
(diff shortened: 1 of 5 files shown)


src/apps/containers.clj
```diff
@@ -585,34 +599,210 @@
...
+                       (where {:container_settings_id (uuidify settings-uuid)
+                               :id                    (uuidify port-uuid)})))))
+
+(defn add-port
+  [settings-uuid port-map]
+  (let [host-port      (:host_port port-map)
+        container-port (:container_port port-map)
+        bind-to-host   (:bind_to_host port-map)]
+    (if (port-mapping? settings-uuid host-port container-port bind-to-host)
+      (throw (Exception. (str "port mapping already exists: " settings-uuid " " host-port " " container-port " " bind-to-host))))
```
slr71: I think this will cause apps to return a 500 status, but a 400 is probably what you want. There's a set of helper functions in `clojure-commons.exception-util` that may help. (I normally use `cxu` as tis namespace alias.) In this case, you can call `cxu/exists` to have the service return a 400 status code along with a response body containing an error code of `ERR_EXISTS`:

```clojure
    (if (port-mapping? settings-uuid host-port container-port bind-to-host)
      (cxu/exists (str "port mapping already exists: " settings-uuid " " host-port " " container-port " " bind-to host)))
```

src/apps/containers.clj
```diff
@@ -585,34 +599,210 @@
...
+(defn tool-port
+  "Returns a map containing port information."
+  [tool-uuid port-uuid]
+  (when (tool-has-settings? tool-uuid)
+    (let [settings-uuid (tool-settings-uuid tool-uuid)]
+      (when (settings-has-port? settings-uuid port-uuid)
+        (dissoc (port port-uuid) :container_settings_id)))))
+
+(defn add-tool-port
+  [tool-uuid port-map]
```
slr71: Destructuring might be helpful in this function as well.

src/apps/containers.clj
```diff
@@ -585,34 +599,210 @@
...
+      (throw (Exception. (str "port mapping already exists: " settings-uuid " " host-port " " container-port " " bind-to-host))))
+    (insert container-ports
+            (values (merge
+                     (select-keys port-map [:host_port :container_port :bind_to_host])
+                     {:container_settings_id (uuidify settings-uuid)})))))
+
+(defn modify-port
+  [settings-uuid port-uuid port-map]
+  (if-not (port? port-uuid)
+    (throw (Exception. (str "port does not exist: " port-uuid))))
```
slr71: You can use `cxu/not-found` in this case.

src/apps/containers.clj
```diff
@@ -585,34 +599,210 @@
...
+  [tool-uuid port-uuid]
+  (when (tool-has-settings? tool-uuid)
+    (let [settings-uuid (tool-settings-uuid tool-uuid)]
+      (when (settings-has-port? settings-uuid port-uuid)
+        (dissoc (port port-uuid) :container_settings_id)))))
+
+(defn add-tool-port
+  [tool-uuid port-map]
+  (when-not (tool-has-settings? tool-uuid)
+    (throw (Exception. (str "Tool " tool-uuid " does not have a container."))))
```
slr71: I think that either `cxu/not-found` or `cxu/bad-request` could be used here, depending on the type of message you want to return.

src/apps/containers.clj
```diff
@@ -585,34 +599,210 @@
...
+  (:interactive_apps_proxy_settings_id
+   (first (select container-settings
+                  (where {:id (uuidify container-settings-uuid)})))))
+
+(defn modify-proxy-settings
+  "Modifies an existing set of proxy settings. Needs the proxy-settings-uuid
+   and a new map of values for the proxy settings."
+  [proxy-settings-uuid proxy-settings]
+  (if-not (proxy-settings? proxy-settings-uuid)
+    (throw (Exception. (str "proxy settings not found:" proxy-settings-uuid))))
```
slr71: `cxu/not-found` could probably be used here.

src/apps/containers.clj
```diff
@@ -585,34 +599,210 @@
...
+        proxy-settings (:interactive_apps info-map)
+        settings       (dissoc info-map :container_devices :container_volumes :container_volumes_from)
+        info-map       (assoc info-map :tools_id (uuidify tool-uuid))]
     (log/warn "adding container information for tool" tool-uuid ":" info-map)
     (transaction
      (let [img-id        (find-or-add-image-id (:image info-map))
            settings-map  (add-settings info-map)
            settings-uuid (:id settings-map)]
        (update-tool {:id tool-uuid :container_images_id img-id})
+       (if-not (nil? proxy-settings)
```
slr71: I like to use `when-not` instead of `if-not` when there's no `else` clause. When I see `if` or `if-not` being used, I tend to look for the else clause even if it's not there.

src/apps/containers.clj
```diff
@@ -621,13 +811,18 @@
           settings    (assoc settings :tools_id tool-id)
           settings-id (:id (add-settings settings))]
       (update-tool {:id tool-id :container_images_id img-id})
-
+      (if-not (nil? interactive_apps)
```
slr71: I'd probably use `when-not` in this case too.

johnworth:
I'd rather not introduce the clojure-commons.exception-util namespace into app.containers as part of this pull request, because I'd have to touch a lot more than just the interactive apps portions of the code in order to keep things consistent.

I've filed https://pods.iplantcollaborative.org/jira/browse/CORE-9218 to track the changes needed for introducing clojure-commons.exception-util.

src/apps/routes/schemas/containers.clj
```diff
@@ -210,17 +210,101 @@
...
+    :container_port                  Integer
+    (s/optional-key :bind_to_host)   Boolean}
+   "Port information for a tool container."))
+
+(s/defschema NewPort
+  (describe
+   (dissoc Port :id)
+   "A map for adding a new port configuration to a tool container."))
+
+(s/defschema HostPort
```
psarando: `HostPort` is not used anywhere. Can we remove it?

src/apps/routes/schemas/containers.clj
```diff
@@ -210,17 +210,101 @@
...
+  (describe
+   {:host_port (s/maybe Integer)}
+   "The port on the host system from which requests are forwarded."))
+
+(s/defschema ContainerPort
```
psarando: `ContainerPort` is not used anywhere. Can we remove it?

src/apps/routes/schemas/containers.clj
```diff
@@ -210,17 +210,101 @@
...
+  (describe
+   {:container_port Integer}
+   "The port that the container is listening on. Info is forwarded to it from the host."))
+
+(s/defschema BindToHost
```
psarando: `BindToHost` is not used anywhere. Can we remove it?

src/apps/routes/schemas/containers.clj
```diff
@@ -210,17 +210,101 @@
...
+    (s/optional-key :ssl_cert_path) String
+    (s/optional-key :ssl_key_path)  String}
+   "Interactive app settings for the reverse proxy that runs on the HTCondor nodes for each job."))
+
+(s/defschema NewProxySettings
+  (describe
+   (dissoc ProxySettings :id)
+   "A map for adding new interactive app reverse proxy settings."))
+
+(s/defschema ProxySettingsImage
```
psarando: I'm only finding refereces to `ProxySettings` and `NewProxySettings`.
Can we remove these other `ProxySettings*` schemas?

johnworth:
From what I All of the dead interactive apps code should now be gone from apps.routes.schema.containers and apps.containers.

Validated the code was actually unused with a combo of ```lein yagni```, ```git grep```, and testing through Swagger locally.

PR #100, opened 2018-05-18
Add (interapps-url) function for later use in notifications

johnworth:
Sarah will be adding support for an interactive apps analysis type later on, at which point we'll be able to add the URL for the interactive app to the Running job status notifications as appropriate. This pull request adds the (interapps-url) function and the (configs/interapps-base) setting to the apps service, so that the URLs can be generated correctly.

### DIFF
--- a/src/apps/clients/notifications.clj
+++ b/src/apps/clients/notifications.clj
@@ -11,7 +11,9 @@
             [apps.clients.notifications.tool-sharing :as tool-notifications]
             [apps.persistence.jobs :as jp]
             [apps.persistence.tool-requests :as tp]
-            [apps.util.config :as config]))
+            [apps.util.config :as config])
+  (:use [pandect.algo.sha256 :only [sha256]]
+        [cemerick.url :only [url]]))
 
 (def ^:private emailable-job-statuses
   #{jp/completed-status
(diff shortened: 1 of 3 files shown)


src/apps/clients/notifications.clj
```diff
@@ -49,6 +51,18 @@
...
+(defn- interapps-url
+  "Returns the externally accessible URL to the interactive app as a URL from
+   cemerick.url.
+
+   Example usage:
+       (str (interapps-url (url (config/interapps-base)) job-info))
+     Returns:
+       https://abb9730df.cyverse.run"
+  [{host :host :as base} {user-id :user_id analysis-id :uuid}]
+  (assoc base :host (str "a" (-> (str user-id analysis-id) sha256 (subs 0 8)) "." host)))
```
psarando: Is prepending "a" to the sha256 required?

johnworth: Yup. The subdomain gets added as a host for a Kubernetes Ingress, which throws an error if the host doesn't start with a letter. I was originally prepending "app-" to the subdomain, but after talking with Sriram and Sarah I decided to shorten it even further.

PR #101, opened 2018-05-24
Parinfer reformatting

johnworth:
I've started using [parinfer](https://github.com/shaunlebron/parinfer) inside Atom as a kind of real-time gofmt for Clojure. It had a bunch of suggested reformatting changes to make which made sense to me.

This should only contain whitespace changes.

### DIFF
--- a/src/apps/routes/admin.clj
+++ b/src/apps/routes/admin.clj
@@ -96,16 +96,16 @@
     :return AdminAppListing
     :description
     (str
-"This service allows admins to list all public apps, including apps listed under the `Trash` category:
+     "This service allows admins to list all public apps, including apps listed under the `Trash` category:
  deleted public apps and private apps that are 'orphaned' (not categorized in any user's workspace).
  If the `search` parameter is included, then the results are filtered by the App name, description,
  integrator's name, tool name, or category name the app is under."
-(get-endpoint-delegate-block
-  "metadata"
-  "POST /avus/filter-targets")
-(get-endpoint-delegate-block
-  "metadata"
-  "POST /ontologies/{ontology-version}/filter-targets"))
+     (get-endpoint-delegate-block
+       "metadata"
+       "POST /avus/filter-targets")
+     (get-endpoint-delegate-block
+       "metadata"
+       "POST /ontologies/{ontology-version}/filter-targets"))
     (ok (coerce! AdminAppListing
                  (apps/admin-search-apps current-user params))))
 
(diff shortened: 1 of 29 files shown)


src/apps/routes/admin.clj
```diff
@@ -143,25 +143,25 @@
       :return AdminAppDetails
       :summary "Update App Details and Labels"
       :description (str
-"This service is capable of updating high-level information of an App,
+                    "This service is capable of updating high-level information of an App,
```
psarando: I think either all of these `:description` strings should remain un-indented (they were imitating HEREDOC style strings), or all of them should be indented at the same level.

The problem is that they can't be indented as they are (as a single string), because Markdown will interpret the leading spaces as a code-block. None of the `:description` strings changed in this PR appear to require newlines where they are, so it just looks like they were formatted this way for readability. But if we indent everything, we have to make them into separate strings passed to `str`:
```clojure
        :description (str
                      "This service is capable of updating high-level information of an App,"
                      " including 'deleted' and 'disabled' flags, as well as just the labels within a single-step app that has"
                      ...
```
etc.

If we go with that format, it wouldn't have to be done for this PR, and it could be updated in a follow-up PR.

johnworth:
On second thought, using parinfer for reformatting a large amount of code probably wasn't the smart thing to do.

src/apps/routes/admin.clj
slr71: I have mixed feelings about this. On one hand, I kind of like the idea of using heredoc strings. On the other hand, I _really_ like letting my text editor handle the formatting without me having to go back and reformat large chunks of code by hand. Since Clojure doesn't have a dedicated heredoc format, I find myself working against my editor when I'm updating the endpoint documentation, especially in cases like this.

I haven't thought too much about it, but it'd be nice if we had a template for endpoint documentation that can be stored in a separate file so that we're not working against the editor for things like this. Alternately, Clojure.core.incubator has a `clojure.core.strint` namespace that implements string interpolation. Maybe that would help us to keep the heredoc emulation without having to fight with our text editors too much.

psarando: I'm also leaning towards letting the IDE handle the formatting.

That was part of the motivation for the `get-endpoint-delegate-block` function, which will add appropriate newlines and such required for Markdown formatting those "sections".

That's also why I suggested in my last comment how we could update these strings to let them be formatted by the IDE consistently. But we'd still have to be careful when formatting other endpoint doc strings that require special newlines and indents for Markdown (but I can only think of 1 data-info endpoint that requires this for now). I haven't checked out `clojure.core.strint`, but that might be a good solution also 👍

PR #102, opened 2018-06-05
CORE-9230: added support for the interactive job type.

slr71:
This required a bunch of little scattered changes, and it wasn't easy to determine which parts of the code had to be changed.


src/apps/persistence/tools.clj
```diff
@@ -32,10 +33,11 @@
...
 
-(defn- get-tool-type-id
-  "Gets the ID of the given tool type name."
-  [tool-type]
-  (:id (first (select tool_types (fields :id) (where {:name tool-type})))))
+(defn- get-tool-type-id-for-tool
+  "Gets the tool type ID to use for a tool."
+  [{:keys [interactive type]}]
+  (when-let [tool-type (if interactive c/interactive-tool-type type)]
+    (:id (first (select tool_types (fields :id) (where {:name tool-type}))))))
```
slr71: This change was intended to ensure that the interactive flag and the tool type are consistent.

src/apps/service/apps/combined/jobs.clj
```diff
@@ -32,8 +28,7 @@
        (partition-by app-step-partitioner)
        (map first)
        (map (fn [n step] (assoc step :step_number n))
-            (iterate inc 1))
-       (map add-job-step-type)))
```
slr71: The query now includes the job step type, so this transformation is no longer required.

src/apps/service/apps/de/jobs.clj
```diff
@@ -15,22 +16,10 @@
...
-  [{{step-type :type} :component :as step}]
-  (if (= step-type "fAPI")
-    (dissoc-in step [:config :input])
-    step))
-
-(defn- pre-process-jex-submission
-  "Finalizes the job for submission to the JEX."
-  [job]
-  (update-in job [:steps] (partial map pre-process-jex-step)))
-
```
slr71: Removing support for fAPI jobs means that we no longer have to do any step preprocessing. 🎉

src/apps/service/apps/de/jobs.clj
```diff
@@ -51,6 +40,13 @@
              :parent_id          (:parent_id submission))
       (jp/save-job submission)))
 
+(defn- job-type-for-step
+  "Determines the job type to use for a single job step."
+  [{{tool-type :type} :component}]
+  (if (= tool-type c/interactive-tool-type)
+    jp/interactive-job-type
+    jp/de-job-type))
```
slr71: For de-only apps, we no longer have access to the app information by the time we get here. The choice was either to ensure that the app information was available here or to make the determination again based on the tool type. I chose the latter for the time being.

src/apps/service/apps/de/jobs/fapi.clj
```diff
@@ -1,70 +0,0 @@
-(ns apps.service.apps.de.jobs.fapi
```
slr71: Let the code deletion begin! 🎉  🎊

slr71:
Thanks for the review!

PR #104, opened 2018-06-05
CORE-9221: added the container entrypoint and image URL to tool listings

slr71:


slr71:
Yeah, that makes sense. I think we'll be okay as long as we make sure to avoid changes that require an additional query for each tool in the listing, but if we run into a case where that's no longer possible or formatting the response body becomes too onerous, then updating the UI will definitely be better.

PR #106, opened 2018-06-06
CORE-9240: added the access URL to job status update notifications fo…

slr71:
…r interactive apps

We need the job step information both in order to determine whether the current job step is interactive and to actually build the access URL. To minimize the scope of this change, I chose to omit cases where it would be cumbersome to obtain the job step information. Fortunately, these cases were all for newly submitted apps as far as I can tell. The upshot is that the access URL will not be included in notifications for interactive apps in the `Submitted` state. This seemed reasonable to me because the URL won't be reachable yet anyway.

### DIFF
--- a/src/apps/clients/notifications.clj
+++ b/src/apps/clients/notifications.clj
@@ -57,11 +57,11 @@
    cemerick.url.
 
    Example usage:
-       (str (interapps-url (url (config/interapps-base)) job-info))
+       (str (interapps-url (url (config/interapps-base)) username external-id))
      Returns:
        https://abb9730df.cyverse.run"
-  [{host :host :as base} {user-id :user_id analysis-id :uuid}]
-  (assoc base :host (str "a" (-> (str user-id analysis-id) sha256 (subs 0 8)) "." host)))
+  [{host :host :as base} username external-id]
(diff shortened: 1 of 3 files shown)


src/apps/clients/notifications.clj
```diff
@@ -57,11 +57,11 @@
...
 
    Example usage:
-       (str (interapps-url (url (config/interapps-base)) job-info))
+       (str (interapps-url (url (config/interapps-base)) username external-id))
      Returns:
        https://abb9730df.cyverse.run"
-  [{host :host :as base} {user-id :user_id analysis-id :uuid}]
-  (assoc base :host (str "a" (-> (str user-id analysis-id) sha256 (subs 0 8)) "." host)))
+  [{host :host :as base} username external-id]
+  (assoc base :host (str "a" (-> (str username external-id) sha256 (subs 0 8)) "." host)))
```
slr71: I made a slight change to this function because the job information available in the apps service is slightly different from the job information available to condor-launcher. It's confusing, but unfortunately, it's easier to change how we get the information than it is to change the available job information at this time.

slr71:
Merging this one now...

PR #108, opened 2018-06-12
CORE-9240: don't include the access URL for completed interactive apps

slr71:


slr71:
Thanks for the review!

PR #109, opened 2018-06-14
CORE-9244: add the OSG tool type, job type and system ID.

slr71:
There's still some testing left to be done, but the testing will involve job submissions. It's probably easiest to merge the changes and test them in the development environment.


slr71:
Thanks for the review!

PR #110, opened 2018-06-18
CORE-9245: allow app listings to be filtered by app type

slr71:


slr71:
Thanks for the review!

PR #111, opened 2018-06-20
CORE-9083: add app integrator info to Agave app listings.

slr71:


slr71:
Thanks for the review!

PR #113, opened 2018-07-20
CORE-4632: modified the app definition retrieval endpoints to use val…

slr71:
…idator argument definitions to determine how to convert argument values


slr71:
Thanks for the review!

PR #115, opened 2018-07-24
CORE-9218: use clojure-commons.exception-util in places were java.lan…

slr71:
…g.Exception was being thrown in apps.containers


slr71:
Thanks or the reviews!!!

PR #116, opened 2018-07-25
CORE-6903: change the notification type for URL upload jobs

slr71:
This may not be the last PR associated with this change. Some notification format changes might be necessary in cases where the notification type is `data`. I want to get this change deployed in our development environment for testing, though.


slr71:
Thanks for the review!

PR #117, opened 2018-09-04
CORE-9280 Add interactive URLs to the analysis listings

ash3rz:
It _seems_ to work! I'm sure there are some clojure tricks and shortcuts I'm unaware of that could be utilized though 😄

### DIFF
--- a/src/apps/routes/schemas/analysis/listing.clj
+++ b/src/apps/routes/schemas/analysis/listing.clj
@@ -66,14 +66,14 @@
    (describe UUID "The identifier of the parent analysis.")
 
    (optional-key :batch_status)
-   (describe BatchStatus "A summary of the status of the batch.")})
+   (describe BatchStatus "A summary of the status of the batch.")
+
+   (optional-key :interactive_urls)
+   (describe [String] "The list of externally accessible interactive analysis URLs.")})
 
 (defschema AdminAnalysis
   (assoc BaseAnalysis
-    :external_ids (describe [ExternalId] "The list of external identifiers.")
-
-    (optional-key :interactive_urls)
-    (describe [String] "The list of externally accessible interactive analysis URLs.")))
+    :external_ids (describe [ExternalId] "The list of external identifiers.")))
 
 (defschema AdminAnalysisList
   {:analyses (describe [AdminAnalysis] "The list of anlayses.")})
@@ -84,10 +84,7 @@
     (describe Boolean "Indicates whether the app is currently disabled.")
 
     :can_share
-    (describe Boolean "Indicates whether or not the analysis can be shared.")
-
-    (optional-key :interactive_urls)
-    (describe [String] "The list of externally accessible interactive analysis URLs.")))
+    (describe Boolean "Indicates whether or not the analysis can be shared.")))
 
 (defschema AnalysisList
   {:analyses  (describe [Analysis] "The list of analyses.")
--- a/src/apps/routes/schemas/analysis/listing.clj
+++ b/src/apps/routes/schemas/analysis/listing.clj
@@ -70,7 +70,10 @@
 
 (defschema AdminAnalysis
   (assoc BaseAnalysis
-    :external_ids (describe [ExternalId] "The list of external identifiers.")))
+    :external_ids (describe [ExternalId] "The list of external identifiers.")
+
+    (optional-key :interactive_urls)
+    (describe [String] "The list of externally accessible interactive analysis URLs.")))
 
 (defschema AdminAnalysisList
   {:analyses (describe [AdminAnalysis] "The list of anlayses.")})
@@ -81,7 +84,10 @@
     (describe Boolean "Indicates whether the app is currently disabled.")
 
     :can_share
-    (describe Boolean "Indicates whether or not the analysis can be shared.")))
+    (describe Boolean "Indicates whether or not the analysis can be shared.")
+
+    (optional-key :interactive_urls)
+    (describe [String] "The list of externally accessible interactive analysis URLs.")))
 
 (defschema AnalysisList
   {:analyses  (describe [Analysis] "The list of analyses.")
--- a/src/apps/service/apps/job_listings.clj
+++ b/src/apps/service/apps/job_listings.clj
@@ -1,11 +1,15 @@
 (ns apps.service.apps.job-listings
-  (:use [kameleon.uuids :only [uuidify]]
-        [apps.util.conversions :only [remove-nil-vals]])
+  (:use [cemerick.url :only [url]]
+        [kameleon.uuids :only [uuidify]]
+        [apps.clients.notifications :only [interapps-url]]
+        [apps.util.conversions :only [remove-nil-vals]]
+        [apps.validation :only [get-valid-user-id]])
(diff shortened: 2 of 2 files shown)


src/apps/service/apps/job_listings.clj
```diff
@@ -68,18 +72,31 @@
     :batch           (:is_batch job)
     :batch_status    (when (:is_batch job) (format-batch-status id))}))
 
+(defn- interactive-urls
+  [job rep-steps]
+  (let [user-id (get-valid-user-id (job :username))
```
psarando: Typically it's safer to access a map's value with `(:key map)`, in case the map is `nil`.
Using a `nil` map as the function will throw an exception, but using the `:key` as the function with a `nil` argument will not.

src/apps/routes/schemas/analysis/listing.clj
```diff
@@ -81,7 +84,10 @@
     (describe Boolean "Indicates whether the app is currently disabled.")
 
     :can_share
-    (describe Boolean "Indicates whether or not the analysis can be shared.")))
+    (describe Boolean "Indicates whether or not the analysis can be shared.")
+
+    (optional-key :interactive_urls)
```
psarando: It looks like this `:interactive_urls` param could go into `BaseAnalysis`, rather than duplicated in `AdminAnalysis` and `Analysis`?

ash3rz:
Thanks for the review!

src/apps/service/apps/job_listings.clj
```diff
@@ -68,18 +72,31 @@
...
     :batch_status    (when (:is_batch job) (format-batch-status id))}))
 
+(defn- interactive-urls
+  [job rep-steps]
+  (let [user-id (get-valid-user-id (:username job))
+        job-id  (:id job)
+        steps   (rep-steps job-id)]
+    (reduce (fn [urls step]
+              (when (= (:job_type step) jp/interactive-job-type)
+                (conj urls (str (interapps-url (url (config/interapps-base)) user-id (:external_id step))))))
```
slr71: I don't think this will be a problem because all of our interactive apps currently only have one step, but I thought I'd mention it just in case. The reduction function currently clears the list of URLs whenever it encounters a non-interactive job step, so this function will always return only the consecutive interactive URLs at the end of the list of job steps. To get a better idea of what's happening, you can take a quick look at this testing I did in the REPL:

```
user=> (reduce (fn [acc n] (when (even? n) (conj acc n))) [] [1 2 3 4 5])
nil
user=> (reduce (fn [acc n] (when (even? n) (conj acc n))) [] [1 2 3 4])
(4)
user=> (reduce (fn [acc n] (when (even? n) (conj acc n))) [] [2 4 6 8])
[2 4 6 8]
user=> (reduce (fn [acc n] (when (even? n) (conj acc n))) [] [1 2 3 4 6 8])
(8 6 4)
user=> (reduce (fn [acc n] (when (even? n) (conj acc n))) '() '(1 2 3 4 6 8))
(8 6 4)
user=> (reduce (fn [acc n] (if (even? n) (conj acc n) acc)) [] [1 2 3 4 6 8])
[2 4 6 8]
```

You can fix the reduction function by having it return the list of URLs if the step that it's currently processing is not interactive:

```clojure
(defn- interactive-urls
  [job rep-steps]
  (let [user-id (get-valid-user-id (:username job))
        job-id  (:id job)
        steps   (rep-steps job-id)]
    (reduce (fn [urls step]
              (if (= (:job_type step) jp/interactive-job-type)
                (conj urls (str (interapps-url (url (config/interapps-base)) user-id (:external_id step))))
                urls))
            []
            steps)))
```

If you wanted to get fancy, you could use `map` and `filter` instead:

```clojure
(defn- interactive-urls
  [job rep-steps]
  (let [user-id      (get-valid-user-id (:username job))
        interactive? (fn [step] (= (:job-type step) jp/interactive-job-type))
        get-url      (fn [step] (str (interapps-url (url (config/interapps-base)) user-id (external_id step))))]
    (map get-url (filter interactive? (rep-steps (:id job))))))
```

I defined a couple of local functions just for readability, but that's completely optional.

The usual caveat that my example code hasn't been tested applies. 😄

src/apps/service/apps/job_listings.clj
```diff
@@ -104,7 +121,9 @@
      :total     (count-jobs user params types analysis-ids)}))
 
 (defn admin-list-jobs-with-external-ids [external-ids]
-  {:analyses (mapv format-admin-job (jp/list-jobs-by-external-id external-ids))})
+  (let [jobs      (jp/list-jobs-by-external-id external-ids)
+        rep-steps (group-by (some-fn :parent_id :job_id) (jp/list-representative-job-steps (mapv :id jobs)))]
```
slr71: This is another thing that probably won't cause a problem because (IIRC) the DE disallows the use of interactive apps in HT batches. With `rep-steps` grouped by `(some-fn :parent_id :job_id)`, however, the result would be that the parent job in a batch would have the URLs of all of the child jobs associated with it, but none of the child jobs would have any URLs associated with them. In this case, I think you can get away with simply grouping by `:job_id` in the same way that `list-job` does.

src/apps/service/apps/job_listings.clj
psarando: Good catch. I overlooked the `when` inside the reducer.
I think you'd typically want the reducer to always return the results so far, but `when` returns `nil` if the condition fails, so that's basically discarding the results-so-far when the check fails.

I think the `if` check makes more sense in the `reduce`, but I also like the `map` + `filter` solution too.

ash3rz: Ooh, goodies to try!  I did not know that about reduce - good catch. I did try my own filter + map combo, but then I noticed for completely non-interactive jobs, I ended up with the `interactive_urls` key with an empty array and I wasn't sure if that was ideal. In my (broken) solution, the key wasn't there.  That being said, I haven't tried your fancy solution, and I like fancy - I'll give it a go!

src/apps/service/apps/job_listings.clj
ash3rz: Ah, another good catch.  I think we did decide we wouldn't allow HT batches, but I want this to be correct in case we do.  I'll update it - thanks!

src/apps/service/apps/job_listings.clj
slr71: Goodies to try are always fun, and Clojure provides a lot of them. I love that about Clojure. :)

Both of my suggestions will result in an empty array under `:interactive_urls`, but you can fix that using `seq`, which converts any collection into a sequence and any empty collection into nil:

```clojure
user=> (seq [])
nil
user=> (seq [1 2 3])
(1 2 3)
user=> (seq '())
nil
user=> (seq {})
nil
user=> (seq {:foo :bar})
([:foo :bar])
user=> (seq '(1 2 3))
(1 2 3)
```

Probably the easiest option would be to put `seq` around the call to `reduce` or `map` depending on which option you choose.

ash3rz: 🌟 Fancy + seq works!  Thanks for taking the time to explain it all!  I'll push the changes shortly

PR #118, opened 2018-09-05
CORE-9308: modified the job status update endpoints to ignore updates…

slr71:
… for completed or canceled jobs


slr71:
Thanks for the reviews!

PR #119, opened 2018-09-06
CORE-9310: process all status updates for a single DE job step at once

slr71:


src/apps/service/apps.clj
```diff
@@ -210,16 +210,28 @@
...
+             job         (jobs/lock-job job-id)
+             batch       (when-let [parent-id (:parent_id job)] (jp/get-job-by-id parent-id))
+             apps-client (get-apps-client-for-username (:username job))]
+         (doseq [{:keys [status sent-on]} updates]
+           (let [end-date (when (jp/completed? status) (str sent-on))
+                 job-step (jobs/get-unique-job-step external-id)]
+             (time (when (jp/status-follows? status (:status job-step))
+                     (jobs/update-job-status apps-client job-step job batch status end-date)))))
+         (jp/mark-job-status-updates-propagated (mapv :id updates))
+         nil))))
```
slr71: The arity used for DE jobs is vastly more complex now. 😖 This should probably be refactored at some point, but I wanted to get this change in as soon as I could. The steps are to get the list of status updates then process each of them in order of the `sent_on` timestamp. After the updates are processed, we'll mark the job status updates as propagated.

Heh, the `time` call should probably be removed. I'll remove it before merging the changes.

src/apps/service/apps.clj
```diff
@@ -210,16 +210,28 @@
...
-            job      (jobs/lock-job job-id)
-            batch    (when-let [parent-id (:parent_id job)] (jobs/lock-job parent-id))]
-        (-> (get-apps-client-for-username (:username job))
-            (jobs/update-job-status job-step job batch status end-date))))))
+   (transaction
+    (let [job-step (jobs/lock-job-step job-id external-id)
+          job      (jobs/lock-job job-id)
+          batch    (when-let [parent-id (:parent_id job)] (jp/get-job-by-id parent-id))]
+      (-> (get-apps-client-for-username (:username job))
+          (jobs/update-job-status job-step job batch status end-date))))))
```
slr71: I allowed Emacs to reindent this code. The only significant change is that I'm no longer locking the batch. Because locking the batch for every child job that is updated is likely to cause significant delays. We may run into concurrency issues because of this. If we do, then the best bet would probably be to lock the batch after we've determined that the batch actually needs to be updated.

src/apps/service/apps.clj
```diff
@@ -210,16 +210,28 @@
...
+   (let [{job-id :job_id} (jobs/get-unique-job-step external-id)
+         updates          (jp/get-job-status-updates external-id)]
+     (when (seq updates)
+       (let [job-step    (jobs/lock-job-step job-id external-id)
+             job         (jobs/lock-job job-id)
+             batch       (when-let [parent-id (:parent_id job)] (jp/get-job-by-id parent-id))
+             apps-client (get-apps-client-for-username (:username job))]
+         (doseq [{:keys [status sent-on]} updates]
+           (let [end-date (when (jp/completed? status) (str sent-on))
+                 job-step (jobs/get-unique-job-step external-id)]
```
ash3rz: Not sure if important, but `(jobs/get-unique-job-step external-id)` is also called on line 214.  Not sure if that previous data can be reused.

slr71: I wanted to get the most recent status of the current job step, and I figured that querying the database again would be much easier and far less error-prone than updating the existing map.

slr71:
I'm going to merge this now. Reviews are still welcome.

slr71:
Thanks for the reviews!

src/apps/service/apps.clj
```diff
@@ -210,16 +210,28 @@
     (format-job-submission-response job-info)))
 
 (defn update-job-status
-  ([external-id status end-date]
-     (let [{job-id :job_id} (jobs/get-unique-job-step external-id)]
-       (update-job-status job-id external-id status end-date)))
+  ([external-id]
```
ianmcorvidae: Just looking at this now, but should this whole block be in a transaction? (or, will it be already from somewhere else that I missed, heh)

slr71: D'oh! Yes it should. Thanks for catching that.

PR #120, opened 2018-09-07
CORE-9306: allow setting a specific status code when stopping jobs

ianmcorvidae:
I still need to test this more than I have, but I wanted to get it up for review. I've also made it so when we stop jobs we put an entry in the `job_status_updates` table, though we do so after setting the real status. I haven't yet evaluated if I should force a propagation of job status updates at the same time, to reduce load elsewhere.

### DIFF
--- a/src/apps/service/apps.clj
+++ b/src/apps/service/apps.clj
@@ -286,7 +286,7 @@
 
 (defn stop-job
   [user job-id params]
-  (let [status (:job_status params "Canceled")]
+  (let [status (:job_status params jp/canceled-status)]
     (jobs/stop-job (get-apps-client user) user job-id status))
   {:id job-id})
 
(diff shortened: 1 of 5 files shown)


src/apps/service/apps.clj
```diff
@@ -285,8 +285,9 @@
   (jobs/get-job-relaunch-info (get-apps-client user) user job-id))
 
 (defn stop-job
-  [user job-id]
-  (jobs/stop-job (get-apps-client user) user job-id)
+  [user job-id params]
+  (let [status (:job_status params "Canceled")]
```
ianmcorvidae: note to self, this should probably use the names in the persistence namespace (maybe the enum above too)

ianmcorvidae:
After some additional testing, this appears functional, so I'll go ahead and merge it, but I'd still love a lookover from @slr71 or @psarando

ianmcorvidae:
(and thanks Ashley for the review!)

Issue #124, opened 2018-10-12
Update /apps/{app-id}/metadata endpoints to support community admin app-tagging


psarando:
Once an app is published, only a community admin should be able to add or remove an app from an App Community.

DE admins should also be able to add or remove an app from an App Community (via /admin endpoints).

PR #127, opened 2018-10-24
Add (/admin)/apps/{app-id}/communities endpoints.

psarando:
Adds endpoints for adding or removing only App Community tags for #124.

### DIFF
--- a/src/apps/service/apps/communities.clj
+++ b/src/apps/service/apps/communities.clj
@@ -1,0 +1,71 @@
+(ns apps.service.apps.communities
+  (:require [apps.clients.iplant-groups :as groups]
+            [apps.clients.metadata :as metadata-client]
+            [apps.util.config :as config]
+            [cheshire.core :as json]
+            [clojure.set :as sets]
+            [clojure-commons.exception-util :as exception-util]))
+
+(defn- get-community-admin-set
+  [username name]
+  (->> (groups/get-community-admins username name)
+       :members
+       (mapv :id)
+       set))
+
+(defn- validate-community-admin
+  [username name]
+  (when-not (contains? (get-community-admin-set username name) username)
+    (exception-util/forbidden "User is not an admin of that community"
+                              :user      username
+                              :community name)))
+
+(defn- validate-avu-community-admins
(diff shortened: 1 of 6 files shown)


src/apps/service/apps/communities.clj
```diff
@@ -0,0 +1,71 @@
...
+  [username app-id {:keys [avus] :as request} admin?]
+  (when-not admin?
+    (validate-avu-community-admins username avus))
+  (metadata-client/update-avus username app-id (json/encode request)))
+
+(defn- metadata->avu-only
+  [avu]
+  (select-keys avu [:attr :value :unit]))
+
+(defn- community-admin-remove-avus
```
psarando: This function currently has to lookup all current AVUs, remove the community AVUs given in the request, then submit the remaining AVUs back to the metadata service with the `PUT /avus/apps/{target-id}` endpoint.

This could cause race-conditions for an app's set/update metadata requests. So I'm thinking it might worth it to add a metadata service `DELETE /avus/{target-type}/{target-id}` endpoint 🤔 

Currently Belphegor has similar behavior when categorizing apps, since it also uses the `PUT /avus/apps/{target-id}` endpoint for ontology category metadata.

ianmcorvidae: We should absolutely add a more specific delete endpoint, probably either what you suggest, with a request body or query parameters to specify some sort of filter, or could be something like `DELETE /avus/{target-type}/{target-id}/{attribute}/{value}/{unit}` (with value/unit optional, probably). I kinda like the latter, but the former might be more flexible since it doesn't impose expectations (i.e., it could let you delete anything with a particular value, or unit, without an attribute filter), and a query parameter might allow fancier filters like regexes maybe too?

slr71: I agree that a more specific delete endpoint might be in order. I tend to like the idea of using query parameters over path parameters because of the added flexibility, but using path parameters might be flexible enough in this case. In other words, I agree with Ian on this one. 😄

psarando:
Thanks for the reviews! 👍

Issue #129, opened 2018-10-30
The job status update endpoint should mark updates for non-existent jobs as propagated.


slr71:
Because we have an externally facing job status update service now, it's possible for job status updates for non-existent jobs to be recorded now. This has happened in some cases where app integrators were testing new OSG apps. This normally doesn't do any harm, but it does cause the status updates to be repeatedly propagated to the apps service, causing unnecessary work.

If the job status update endpoint receives an update for a job that doesn't exist, it should mark every job status update with that job ID as propagated in order to prevent additional propagation attempts.

PR #130, opened 2018-10-30
#129: mark job status updates for non-existent job steps as propagated …

slr71:
…if there are too many or too few matching job steps for processing

### DIFF
--- a/src/apps/service/apps.clj
+++ b/src/apps/service/apps.clj
@@ -219,12 +219,11 @@
 
 (defn update-job-status
   ([external-id]
+   (jobs/validate-job-status-update-step-count external-id)
    (transaction
-    (let [updates (jp/get-job-status-updates external-id)
-          steps   (jp/get-job-steps-by-external-id external-id)]
+    (let [updates (jp/get-job-status-updates external-id)]
       (when (seq updates)
-        (jobs/validate-job-status-update-step-count external-id updates steps)
-        (let [job-id      (:job_id (first steps))
+        (let [job-id      (:job_id (jp/get-job-steps-by-external-id external-id))
               job-step    (jobs/lock-job-step job-id external-id)
               job         (jobs/lock-job job-id)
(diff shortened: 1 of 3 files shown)


src/apps/service/apps.clj
```diff
@@ -219,17 +219,18 @@
...
 (defn update-job-status
   ([external-id]
+   (jobs/validate-job-status-update-step-count external-id)
    (transaction
-    (let [{job-id :job_id} (jobs/get-unique-job-step external-id)
-          updates          (jp/get-job-status-updates external-id)]
+    (let [updates (jp/get-job-status-updates external-id)]
       (when (seq updates)
-        (let [job-step    (jobs/lock-job-step job-id external-id)
+        (let [job-id      (:job_id (jp/get-job-steps-by-external-id external-id))
```
psarando: It seems like `get-job-steps-by-external-id` will return a list of results?
Should `first` be called on the results before `:job_id`?

src/apps/service/apps/jobs.clj
```diff
@@ -14,16 +14,18 @@
...
-  "Gets a unique job step for an external ID. An exception is thrown if no job step
-  is found or if multiple job steps are found."
+(defn validate-job-status-update-step-count
+  "Validates the number of steps for an external ID provided in a set of job status updates. If there are too many or
+  too few matching job steps then all of the job status updates will be marked as propagated and an exception will be
+  thrown. This function should not be called from within a transaction."
   [external-id]
   (let [job-steps (jp/get-job-steps-by-external-id external-id)]
     (when (empty? job-steps)
+      (jp/mark-job-status-updates-for-external-id-completed external-id)
```
psarando: 👍

src/apps/service/apps.clj
slr71: Ah, yes it should. Thanks for catching that.

slr71:
Thanks for the review!

PR #133, opened 2018-11-09
prevent batch statuses for canceled jobs from reverting

slr71:


slr71:
Thanks for the review!

PR #134, opened 2018-11-21
allow integer and double parameter validation rule values to be handl…

slr71:
…ed correctly


slr71:
Thanks for the review!

PR #135, opened 2018-12-05
added the `for-job` query parameter to the ticket creation and deleti…

slr71:
…on endpoint calls


slr71:
Thanks for the review!

PR #138, opened 2018-12-20
Allow updates for private OSG and Interactive tool types.

psarando:
This PR will allow private tool updates for OSG and Interactive tool types, and also set the network mode to "bridge" for interactive tools types.


psarando:
Thanks for the review 👍

PR #139, opened 2019-01-25
CORE-47: leave the app integrator information unchanged for Agave app…

slr71:
…s if it can't be found


slr71:
Thanks for the review! Merging now...

PR #141, opened 2019-02-15
CORE-5: moved `AppIdParam` to `common-swagger-api.schema.apps`

slr71:


project.clj
```diff
@@ -28,7 +28,7 @@
                  [org.cyverse/metadata-client "3.0.1"]
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/common-cfg "2.8.1"]
-                 [org.cyverse/common-swagger-api "2.8.3"]
+                 [org.cyverse/common-swagger-api "2.9.3-SNAPSHOT"]
```
yeea: Just curious why you've migrated to a SNAPSHOT version (that is, what's the purpose or limitation etc)?

slr71: The only reason I switched to a snapshot version was because I anticipate making more changes, and I didn't want to create a flurry of new releases for the library. I'll create a release version of `common-swagger-api` and modify other projects to use it when we're done adding the Swagger docs.

slr71:
Thanks for the reviews!

project.clj
yeea: Oh, neat! That clears that up for me, thanks!

PR #142, opened 2019-02-22
CORE-105: allow job submissions to work for apps with default inputs

slr71:


src/apps/service/apps/de/jobs/common.clj
```diff
@@ -29,7 +29,7 @@
 
 (defn build-default-values-map
   [params]
-  (remove-vals #(string/blank? (str %))
+  (remove-vals (comp string/blank? str)
```
psarando: 👍

src/apps/service/apps/jobs/submissions.clj
```diff
@@ -207,11 +207,16 @@
...
+  [property-values input-params-by-id]
+  (let [get-property-value (fn [id] (get property-values id (get-in input-params-by-id [id :defaultValue])))]
+    (into {} (map (juxt identity get-property-value) (keys input-params-by-id)))))
+
 (defn submit
   [apps-client user {app-id :app_id system-id :system_id :as submission}]
   (let [[job-types app]    (.getAppSubmissionInfo apps-client system-id app-id)
         input-params-by-id (get-app-params app ap/param-ds-input-types)
-        input-paths-by-id  (select-keys (:config submission) (keys input-params-by-id))
+        input-paths-by-id  (build-input-path-map (:config submission) input-params-by-id)
```
psarando: It looks like the submission config used to contain inputs as `id` keys to `input path` values, but now it will contain `id` keys to `input object` values, and the input object/map will always contain a `defaultValue` key with a path value?

slr71: I'm not entirely sure I understand the question, but the config JSON in the submission hasn't changed. The way that it works now is that the qualified property ID (which is in the format, `{step_id}_{property_id}` is associated with one or more paths, for example:

```clojure
{
    :b22e6469-73a4-452c-9ecb-0094ef43aacc_4775d35c-bdc1-4fdd-92f2-948f900d0c05: "/path/to/some/file",
    :3e9c0c9e-5ee4-4a58-a4c1-edd459381030_7c3cdf20-f62b-47e0-873d-204aedb4d816: [
        "/path/to/some/other/file",
        "/path/to/yet/another/file"
    ]
}
```

The parameter itself may or may not have a default value. For the time being the default value is always a path, although we may want to change it to a UUID in the near future.

Hopefully this answers the question. Let me know if I'm off the mark.

slr71:
Thanks for the review!

src/apps/service/apps/jobs/submissions.clj
psarando: Ah, sorry about that. In my haste to finish this review before the meeting, I misread `(get property-values id (get-in input-params-by-id [id :defaultValue]))` as `(get-in property-values [id :defaultValue])`, so I thought that meant the submission config would always have a `defaultValue` now.

Now I see that we are still pulling the path from the submission config, but if one isn't present, we use the `defaultValue` from the appropriate `input-params-by-id` map instead 😅

slr71: Ah, no worries. I was afraid that there might have been a bug in the code that I'd missed somewhere. 😅

slr71:
Thanks for the review!

PR #143, opened 2019-02-26
CORE-106: added access validation for default inputs.

slr71:


src/apps/service/apps/de/job_view.clj
```diff
@@ -75,24 +99,42 @@
...
       (assoc :label     name
-             :groups    (remove (comp empty? :parameters) (format-steps app-id include-hidden-params?))
+             :groups    (remove (comp empty? :parameters) (format-steps user app-id include-hidden-params?))
              :app_type  "DE"
              :system_id c/system-id)))
 
+(defn- validate-hidden-inputs [user app-id]
+  (when-let [paths (mapv :default_value (filter util/input? (mp/load-hidden-params app-id)))]
+    (try+
+     (data-info/get-path-info user :paths paths :validation-behavior "read")
```
ianmcorvidae: You should add a `:filter-include` here, probably `:filter-include "path"`. By only including that key, you minimize the amount of work that data-info has to do. With just `path` the only validation it'll do is that it exists and is readable, basically, and won't do anything like looking up UUIDs or info-types or etc. See also `apps.service.apps.jobs.util/create-output-dir` which does a similar `get-path-info` in a `try+` in order to check the existence/readability of an object in irods.

slr71: Ah, good idea. Thanks for the suggestion.

src/apps/service/apps/de/job_view.clj
```diff
@@ -31,12 +36,30 @@
                       {:is_implicit false})))
       select))
 
+(defn- path-accessible?
+  [user path]
+  (try+
+   (data-info/get-path-info user :paths [path] :validation-behavior "read" :filter-include "path")
+   (catch [:status 500] e
+     (let [error-code (:error_code (service/parse-json (:body e)))]
+       (if (#{"ERR_NOT_READABLE" "ERR_DOES_NOT_EXIST"} error-code)
```
psarando: 👍

src/apps/service/apps/de/job_view.clj
```diff
@@ -56,16 +79,17 @@
...
 
 (defn- format-group
-  [name-prefix include-hidden-params? step group]
-  {:id          (:id group)
-   :name        (str name-prefix (:name group))
-   :label       (str name-prefix (:label group))
-   :parameters  (mapv (partial format-parameter step) (get-parameters (:id step) (:id group) include-hidden-params?))
-   :step_number (:step_number step)})
+  [user name-prefix include-hidden-params? step group]
+  (let [params (get-parameters (:id step) (:id group) include-hidden-params?)]
```
psarando: 👍

slr71:
Thanks for the reviews!

PR #144, opened 2019-02-27
CORE-122: changed the default value format for input parameters

slr71:


slr71:
Thanks for the review!

PR #145, opened 2019-02-28
CORE-63: Add code to load extra fields when submitting

ianmcorvidae:
This pulls the `extra_requirements` column from `apps_htcondor_extra` and adds it to the submission.

Still needs the downstream jex-adapter etc. work, of course, but hopefully if I did anything too blatantly dumb it can get caught here :laughing:

### DIFF
--- a/src/apps/service/apps/de/jobs/common.clj
+++ b/src/apps/service/apps/de/jobs/common.clj
@@ -114,7 +114,7 @@
        (take-while (comp nil? :external_app_id))
        (reduce #(.buildStep request-builder %1 %2) [])))
 
-(defn load-htcondor-extra-requirements
+(defn- load-htcondor-extra-requirements
   [app-id]
   (first
     (select :apps_htcondor_extra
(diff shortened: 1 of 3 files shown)


src/apps/service/apps/de/jobs/common.clj
```diff
@@ -114,6 +114,16 @@
        (take-while (comp nil? :external_app_id))
        (reduce #(.buildStep request-builder %1 %2) [])))
 
+(defn load-htcondor-extra-requirements
+  [app-id]
+  (select :apps_htcondor_extra
+          (fields [:extra_requirements])
+          (where {:app_id app-id})))
```
ianmcorvidae: ... I've already realized I probably need to add `(first (first ))` or something here

ianmcorvidae:
Cool, thanks for the review! I'll probably merge this after I finish the other work to make these fields... actually at all useful, lol

src/apps/service/apps/de/jobs/common.clj
```diff
@@ -114,6 +114,17 @@
        (take-while (comp nil? :external_app_id))
        (reduce #(.buildStep request-builder %1 %2) [])))
 
+(defn load-htcondor-extra-requirements
```
psarando: I'd make this function private (`defn-`), just to make it obvious it's not in use outside this namespace.

ianmcorvidae: Ah, yeah, good idea.

PR #147, opened 2019-03-13
CORE-11 Migrate /apps listing docs into common-swagger-api.

psarando:
This PR will migrate the initial `/apps` endpoint docs into the common swagger lib, starting with just the `/apps` search endpoint for this PR.

If these initial PRs get approved, then the next PR will probably migrate docs for more than 1 endpoint.


psarando:
Thanks for the review! 👍

PR #148, opened 2019-03-20
CORE-133: modified the code to catch iRODS ticket deletion errors for…

slr71:
… completed or failed jobs


slr71:
Thanks for the review!

PR #149, opened 2019-03-21
CORE-134: set a lock timeout before attempting to obtain locks on a j…

slr71:
…ob or job step

With this change, the following response will be returned when a lock can't be obtained:

```
$ curl -sH "Content-Type: application/json" -d '{"uuid":"2c6e0013-7c62-4f2f-a076-656c61115786"}' "http://localhost:31323/callbacks/de-job" | jq .
{
  "error_code": "ERR_UNCHECKED_EXCEPTION",
  "reason": "org.postgresql.util.PSQLException: ERROR: canceling statement due to lock timeout\n  Where: while locking tuple (14,49) in relation \"job_steps\""
}
```

### DIFF
--- a/src/apps/persistence/jobs.clj
+++ b/src/apps/persistence/jobs.clj
@@ -798,3 +798,8 @@
   (sql/update :job_status_updates
               (set-fields {:propagated true})
               (where {:external_id external-id})))
+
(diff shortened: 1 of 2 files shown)


slr71:
Thanks for the review!

PR #150, opened 2019-03-21
CORE-127: automatically publish tools when an app is being published

slr71:
My biggest concern here is that there might be some more corner cases that I haven't thought of. If reviewers could keep this in mind while looking over the code, I'd appreciate it.

### DIFF
--- a/src/apps/service/apps/de/validation.clj
+++ b/src/apps/service/apps/de/validation.clj
@@ -9,7 +9,9 @@
         [korma.core :exclude [update]]
         [slingshot.slingshot :only [try+ throw+]])
   (:require [apps.clients.permissions :as perms-client]
-            [apps.service.apps.de.permissions :as perms]))
+            [apps.service.apps.de.permissions :as perms]
+            [apps.util.config :as config]
+            [clojure.string :as string]))
 
 (defn validate-app-existence
   "Verifies that apps exist."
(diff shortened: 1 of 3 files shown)


src/apps/service/apps/de/validation.clj
```diff
@@ -45,6 +45,18 @@
...
+(defn- remove-owned-tools
+  [username tools]
+  (let [tool-ids  (mapv :id tools)
+        owned-ids (set (keys (perms-client/load-tool-permissions username tool-ids "own")))]
+    (remove (comp owned-ids :id) tools)))
+
+(defn- get-unpublishable-tools
+  [username tools]
+  (let [public-tool-ids (perms-client/get-public-tool-ids)]
+    (->> (remove (comp public-tool-ids :id) tools)
```
psarando: 👍

src/apps/service/apps/de/validation.clj
```diff
@@ -45,6 +45,18 @@
...
+         (remove-owned-tools username))))
```
psarando: We had also discussed restricting auto-publish only for tools that had a container image that was already in our private registry, so the image name had to start with `docker.cyverse.org/`.

Should we implement this restriction in these service endpoints as well?

slr71: Ah, I'd forgotten about that requirement. Thanks for mentioning it. Yeah, that requirement should be also be in the services.

slr71: The validation has been added. I think this is ready for another review.

slr71:
Thanks for the review!

slr71:
s/review/reviews/ 😛

PR #151, opened 2019-03-22
CORE-11 Migrate /apps/elements docs into common-swagger-api.

psarando:


src/apps/routes/apps/elements.clj
```diff
@@ -2,79 +2,55 @@
...
-        input. For example, if a parameter value must be an integer between 1 and 10 then the
-        `IntRange` rule type may be used. Similarly, if a parameter value must contain data in a
-        specific format, such as a phone number, then the `Regex` rule type may be used."
+        :return elements-schema/RuleTypeListing
+        :summary elements-schema/AppElementsRuleTypeListingSummary
+        :description elements-schema/AppElementsRuleTypeListingDocs
         (ok (list-elements "rule-types" params)))
 
+  ;; Deprecated?
   (GET "/tools" []
```
psarando: We can also mark this endpoint as `deprecated` in the Swagger UI, but the current version of `compojure-api` doesn't support the `deprecated` flag natively.

We could add support with a method like the following, perhaps in `common-swagger-api.routes`:
```clojure
(defmethod compojure.api.meta/restructure-param :deprecated
  [_ deprecated? acc]
  (update-in acc [:swagger] assoc :deprecated deprecated?))
```

Then adding `:deprecated true` to the route will mark it as deprecated in the Swagger UI: https://swagger.io/docs/specification/paths-and-operations/

Of course, if this endpoint is no longer in use, it's probably best to just remove it in this case, but I thought I'd take the opportunity to note the `deprecated` flag here.

psarando:
I'm going to merge this now, but reviews are still welcome 👍

PR #154, opened 2019-04-05
CORE-11 Migrate more schemas and docs from apps.routes.app

psarando:
This PR will migrate the non-admin schemas from `apps.routes.schemas.app` and the endpoint docs from `apps.routes.apps` to `common-swagger-api.schema.apps` (see cyverse-de/common-swagger-api#15).

This requires migrating schemas from the following namespaces:
*  from `apps.routes.schemas.containers` to `common-swagger-api.schema.containers`
* from `apps.routes.schemas.tool` to `common-swagger-api.schema.tools`
* from `apps.routes.schemas.integration-data` to `common-swagger-api.schema.integration-data`

It might be easier to review this PR one commit at a time. Or (as I've done with similar PRs), open this PR in one tab and cyverse-de/common-swagger-api#15 in another, then swap back and forth between tabs as a kind of [blink comparator](https://en.wikipedia.org/wiki/Blink_comparator).


src/apps/routes/apps.clj
```diff
@@ -1,29 +1,86 @@
...
-                                                          AppSharingDocs
-                                                          AppSharingRequest
-                                                          AppSharingResponse
-                                                          AppSharingSummary
-                                                          AppUnsharingDocs
-                                                          AppUnsharingRequest
-                                                          AppUnsharingResponse
-                                                          AppUnsharingSummary]]
+        [common-swagger-api.schema.apps
+         :only [App
```
psarando: I moved the `:only` list to a new line since I thought it looked a little better with these longs lists.
I can reformat these lists with the `:only` on the same line as the namespace if everyone else prefers that style.

Now that I think about it, it might even be better to require a namespace alias (like `common-swagger-api.schema.apps  :as apps-schema`) and use that instead of these long `:use ... :only` lists 🤔

src/apps/routes/admin.clj
```diff
@@ -129,7 +138,7 @@
 
   (context "/:system-id/:app-id" []
     :path-params [system-id :- SystemId
-                  app-id    :- AppIdJobViewPathParam]
+                  app-id    :- StringAppIdParam]
```
psarando: These were the same schema, so I just deleted `AppIdJobViewPathParam` and used `StringAppIdParam` that was already in the common swagger lib.

src/apps/routes/schemas/tool.clj
```diff
@@ -122,13 +77,6 @@
       (->optional-param :implementation)
       (->optional-param :container)))
 
-(defschema NewTool
-  (assoc Tool
-    :id (describe String "The tool's ID")))
-
-(defschema NewToolListing
```
psarando: This is used as the response schema for the `GET /apps/{system-id}/{app-id}/tools` endpoint, so it was renamed to `AppToolListing` in cyverse-de/common-swagger-api#15.

`AppToolListing` is now a list of `AppDetailsTool`, which is the same as the deleted `NewTool` above, except `AppDetailsTool` also includes an optional key for `:container` (which is a `ToolListingImage`).

The `GET /apps/{system-id}/{app-id}/tools` endpoint currently doesn't return the `:container` field, but this key is optional, and now the endpoint could return that info in the listing, if required, without having to update the schema further.

src/apps/routes/apps.clj
psarando: Actually, I've convinced myself that I prefer the `[common-swagger-api.schema.apps :as schema]` alias, so I'll push a commit that removes this `:only` list.

I'm leaving the `common-swagger-api.schema.apps.permission :only` list below, since it's short enough, but I can still update its style if preferred.

psarando:
Thanks for the review! 👍

PR #157, opened 2019-04-11
CORE-185: endpoint to get submission launch info

slr71:
This functionality will eventually be moved to the `analyses` service, but migrating it now would require too much work, and there's little use in duplicating it. The new endpoint is intended to be called by the `analyses` service to get the launch information for a submission.


slr71:
I realized this morning that the code that I reused for this doesn't check the permissions on input files associated with the quick launch. I'm going to close this PR for now until I can implement the permissions checks.

PR #158, opened 2019-04-11
CORE-11 Migrate /apps/hierarchies docs and schemas

psarando:
This PR will use the `/apps/hierarchies` endpoint docs and schemas migrated in cyverse-de/common-swagger-api#18.


psarando:
Merging now, but reviews still welcome 👍

PR #162, opened 2019-04-18
CORE-11 Use common-swagger-api metadata docs and schemas

psarando:
This PR will update `/apps/{app-id}/metadata` endpoint docs to use the AVU schemas and docs added by cyverse-de/common-swagger-api#21.

This also requires the updated metadata-client functions added by cyverse-de/metadata-client#5.


psarando:
I'm going to merge this now, but reviews are still welcome 👍

PR #164, opened 2019-05-01
Submit VICE analyses to K8s

johnworth:
Modifies the apps service to POST VICE analyses to the app-exposer service instead of jex-adapter. Currently disabled through a configuration setting.


src/apps/clients/vice.clj
```diff
@@ -0,0 +1,22 @@
...
+  [uuid]
+  (http/post (app-exposer-url "vice" uuid "save-and-exit")
+             {:throw-exceptions false
+              :as               :json}))
+
+(defn submit-job
+  [job]
+  (http/post (app-exposer-url "vice" "launch")
+             {:content-type :json
+              :body         (cheshire/encode job)}))
```
slr71: You can replace `:body (cheshire/encode job)` with `:form-params job` if you want. clj-http will automatically do the JSON encoding as long as the content type is `:json`. This isn't a requirement or anything, just a convenience that I use a lot.

PR #165, opened 2019-05-01
CORE-11 Migrate remaining /tools endpoint schemas and docs

psarando:
This PR will replace the remaining `/tools` endpoint schemas and docs in `apps.routes.tools` with those added by cyverse-de/common-swagger-api#23.


psarando:

PR #166, opened 2019-05-04
CORE-11 Migrate /analyses schemas and docs

psarando:
This PR will replace the `/analyses` endpoint schemas and docs in `apps.routes.analyses` with those added by cyverse-de/common-swagger-api#24, and also rearrange the order of some of these endpoints.

It will also remove unused `Jex`* schemas from `apps.routes.schemas.analysis`.


psarando:

PR #169, opened 2019-05-10
CORE-21 Move webhook schemas to common-swagger-api

ash3rz:


ash3rz:
Thanks for the reviews!

PR #172, opened 2019-05-18
CORE-11 Migrate schemas and docs for /reference-genomes endpoints

psarando:
This PR will replace the schemas and docs for `/reference-genomes` endpoints with those added by cyverse-de/common-swagger-api#32.


psarando:

PR #176, opened 2019-05-31
CORE-63: Add handling for htcondor extra requirements

ianmcorvidae:
This allows the field to be added in the administrative app PATCH endpoint, and returns it in the administrative app details endpoint.

Depends on cyverse-de/common-swagger-api#36 and cyverse-de/common-swagger-api#37 but the relevant snapshot version was already what was in project.clj of this project, so the changes are only to the functionality of the endpoints themselves.

It may also be useful to add documentation information into the app details endpoint, as we discussed at one point. That hasn't happened yet, but could get added to this PR or go into a new one.

We'll also need to pull the new common-swagger-api version into terrain, but I don't see much need to make a PR for just that :)

### DIFF
--- a/src/apps/persistence/app_metadata.clj
+++ b/src/apps/persistence/app_metadata.clj
@@ -405,6 +405,12 @@
    (delete app_references (where {:app_id app-id}))
    (dorun (map (partial add-app-reference app-id) references))))
 
+(defn set-htcondor-extra
+  [app-id extra-requirements]
+  (transaction
+    (delete apps_htcondor_extra (where {:apps_id app-id}))
+    (insert apps_htcondor_extra (values {:apps_id app-id, :extra_requirements extra-requirements}))))
+
 (defn- get-job-type-id-for-system* [system-id]
   (:id (first (select :job_types (fields :id) (where {:system_id system-id})))))
 
--- a/src/apps/persistence/app_metadata.clj
+++ b/src/apps/persistence/app_metadata.clj
@@ -17,6 +17,7 @@
             [apps.persistence.app-metadata.delete :as delete]
             [apps.persistence.app-metadata.relabel :as relabel]
             [clojure.set :as set]
+            [clojure.string :as string]
             [clojure-commons.exception-util :as cxu]
             [korma.core :as sql]))
 
@@ -409,7 +410,8 @@
   [app-id extra-requirements]
   (transaction
     (delete apps_htcondor_extra (where {:apps_id app-id}))
-    (insert apps_htcondor_extra (values {:apps_id app-id, :extra_requirements extra-requirements}))))
+    (if-not (or (nil? extra-requirements) (empty? (string/trim extra-requirements)))
+      (insert apps_htcondor_extra (values {:apps_id app-id, :extra_requirements extra-requirements})))))
 
 (defn- get-job-type-id-for-system* [system-id]
   (:id (first (select :job_types (fields :id) (where {:system_id system-id})))))
--- a/src/apps/persistence/entities.clj
+++ b/src/apps/persistence/entities.clj
@@ -7,7 +7,7 @@
          parameter_values parameter_types value_type validation_rules validation_rule_arguments
          rule_type rule_subtype app_category_listing app_listing tool_listing ratings collaborators
          genome_reference created_by last_modified_by data_source tool_types
-         tool_request_status_codes tool_architectures tool_requests
+         tool_request_status_codes tool_architectures tool_requests apps_htcondor_extra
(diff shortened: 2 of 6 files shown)


src/apps/persistence/app_metadata.clj
```diff
@@ -405,6 +405,12 @@
    (delete app_references (where {:app_id app-id}))
    (dorun (map (partial add-app-reference app-id) references))))
 
+(defn set-htcondor-extra
+  [app-id extra-requirements]
+  (transaction
+    (delete apps_htcondor_extra (where {:apps_id app-id}))
+    (insert apps_htcondor_extra (values {:apps_id app-id, :extra_requirements extra-requirements}))))
```
ianmcorvidae: I'm not sure if it matters, but I could make it so this `insert` is only done if the extra-requirements are nonempty? The column on the table is only `NOT NULL`, however, and it won't hurt anything to have empty strings in there, so, shrug

src/apps/service/apps/de/listings.clj
```diff
@@ -488,6 +488,12 @@
     (if admin? (jobs-db/get-job-stats app-id params)
                (jobs-db/get-public-job-stats app-id params))))
 
+(defn- format-app-extra-info
+  [app-id admin?]
+  (if admin?
+    (get-app-extra-info app-id)
+    nil))
```
ianmcorvidae: This and the persistence function are both set up so that the entire `extra` key will only be included if there's something there (even an empty string, maybe), but otherwise be totally gone. We could probably change it to always include the key but default to an empty string, if that would be more useful for the UI or such, but I haven't worried about it yet.

src/apps/service/apps/de/admin.clj
```diff
@@ -80,26 +80,35 @@
...
-  [{app-id :id :keys [references groups] :as app}]
+  [{app-id :id :keys [references groups extra] :as app}]
   (persistence/update-app app)
   (when-not (empty? references)
     (persistence/set-app-references app-id references))
   (when-not (empty? groups)
-    (update-app-labels (select-keys app [:id :groups]))))
+    (update-app-labels (select-keys app [:id :groups])))
+  (when-not (empty? extra)
+    (update-app-extra-info app)))
```
psarando: It seems like we don't have a way to delete extra info.

If `set-htcondor-extra` is updated as suggested, so that an `insert` is not attempted if the value is empty, then this `when-not` check could be removed, and the `when-let` checks in `update-app-extra-info` could also be removed to allow a nil to be passed to `set-htcondor-extra` in order to remove the current value.

src/apps/persistence/entities.clj
```diff
@@ -50,6 +50,9 @@
 ;; References associated with an app.
 (defentity app_references)
 
+;; Extra info for HTCondor
+(defentity apps_htcondor_extra)
```
psarando: We probably don't need this `defentity` defined, since we're not depending on korma to join this table in any queries. So `get-app-extra-info` could select using a keyword instead: `select :apps_htcondor_extra (fields ...`

It also doesn't hurt to leave it defined, now that it's already added 👍

src/apps/service/apps/de/admin.clj
ianmcorvidae: I think it'll be sufficient to make the change to `set-htcondor-extra` -- `extra` won't be `empty?` unless there's nothing inside it, i.e.:

```
apps.core=> (empty? {:htcondor {:extra_requirements nil}})
false
apps.core=> (empty? {:htcondor {:extra_requirements ""}})
false
```

so with the change I'm about to push I think it'll be removed when an empty string is passed in, and that makes the endpoint behave more consistently with itself (if a key's that can be patched isn't included, don't touch it at all)

src/apps/persistence/app_metadata.clj
ianmcorvidae: I think it should work now to pass that, or to pass an empty string (or, for good measure, a string that's only spaces).

src/apps/persistence/entities.clj
ianmcorvidae: Yeah, I weighed looking up how to do it with the keyword and just figured it was easier to add them into this namespace. Maybe at some point we'll standardize, but I don't think this is any worse than our existing split, heh.

src/apps/service/apps/de/admin.clj
psarando: cool, that makes sense too.

psarando:
> but the relevant snapshot version was already what was in project.clj of this project

Actually, I think I've bumped the `common-swagger-api` version since then, so I think you'll need to bump the version in apps after those PRs are merged.

PR #177, opened 2019-06-06
CORE-11 Migrate schemas and docs for /admin/tools endpoints

psarando:
This PR will replace schemas and docs for `/admin/tools` endpoints with those added by cyverse-de/common-swagger-api#39.


psarando:

PR #178, opened 2019-06-07
Remove unused /admin/tools/{tool-id}/container endpoints

psarando:
Updates to all tool container settings are handled by the `POST` and `PATCH /admin/tools/{tool-id}` endpoints.


psarando:
Merging now, so the weekend QA tests can hit this, but reviews are still welcome 🙇

PR #180, opened 2019-07-12
CORE-63: Add app documentation to admin app details where available

ianmcorvidae:
Includes update of common-swagger-api for cyverse-de/common-swagger-api#42

### DIFF
--- a/src/apps/service/apps/de/listings.clj
+++ b/src/apps/service/apps/de/listings.clj
@@ -18,6 +18,7 @@
             [apps.persistence.jobs :as jobs-db]
             [apps.service.apps.de.constants :as c]
             [apps.service.apps.de.permissions :as perms]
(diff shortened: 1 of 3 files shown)


src/apps/service/apps/de/listings.clj
```diff
@@ -494,6 +495,13 @@
     (get-app-extra-info app-id)
     nil))
 
+(defn- format-app-documentation
+  [app-id username admin?]
+  (when admin?
+    (try+
+      (docs/get-app-docs username app-id admin?)
+      (catch [:type :clojure-commons.exception/not-found] _ nil))))
```
ianmcorvidae: (this is necessary because the get-app-docs function was written for the docs endpoint rather than for this -- if the whole app itself isn't found, the exception will be thrown elsewhere, but if the docs don't exist we just don't include them)

PR #181, opened 2019-07-17
CORE-324 Update app publish endpoint to only notify community admins.

psarando:
The app publish endpoint will no longer auto-tag apps with communities.

Instead, community admins will now receive notifications that the integrator wishes to tag their public app with those communities.

All other metadata, including ontology tags, are still added to the app.

### DIFF
--- a/src/apps/service/apps/de/metadata.clj
+++ b/src/apps/service/apps/de/metadata.clj
@@ -164,7 +164,7 @@
   (zipmap admin-set (repeat [community-name])))
 
 (defn- notify-community-admins
-  [username app-name community-names]
+  [username integrator-name app-name community-names]
   (->> community-names
        (map #(admin->communities-map % (communities/get-community-admin-set username %)))
        (apply merge-with into)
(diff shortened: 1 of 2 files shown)


src/apps/service/apps/de/metadata.clj
```diff
@@ -173,15 +173,24 @@
...
-       (map (partial apply notifications/send-community-admin-notification username app-name))
+       (map (partial apply notifications/send-community-admin-notification username integrator-name app-name))
        dorun))
 
 (defn- publish-app-metadata
+  "Publishes all metadata sent in the publish request, including ontology AVU tags, but does not publish community tags.
+   Instead, community admins are notified that the app integrator wishes to add their app to those communities."
   [username app-id app-name avus]
-  (let [body (cheshire/encode {:avus (conj avus (beta-avu))})]
+  (let [body (as-> avus m
```
psarando: Inspired by today's chat in CyVerse slack!

psarando:
Thanks for the review! 👍

PR #182, opened 2019-08-08
CORE-305: encode file URLs that are sent to Aloe

slr71:


slr71:
I'm going to merge this change immediately. Reviews are still welcome.

PR #183, opened 2019-08-20
CORE-463 Add total count to tool listing endpoint

ash3rz:
I realized this piece was missing for the tool listing endpoint to be ready for pagination.

### DIFF
--- a/src/apps/tools.clj
+++ b/src/apps/tools.clj
@@ -72,10 +72,14 @@
   [{:keys [user] :as params}]
   (let [public-tool-ids (perms-client/get-public-tool-ids)
         perms           (perms-client/load-tool-permissions user)
-        tool-ids        (filter-listing-tool-ids (set (keys perms)) public-tool-ids params)]
-    {:tools
+        tool-ids        (filter-listing-tool-ids (set (keys perms)) public-tool-ids params)
+        paged-params    (assoc params :tool-ids tool-ids :deprecated false)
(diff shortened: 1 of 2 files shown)


src/apps/tools.clj
```diff
@@ -72,10 +72,14 @@
   [{:keys [user] :as params}]
   (let [public-tool-ids (perms-client/get-public-tool-ids)
         perms           (perms-client/load-tool-permissions user)
-        tool-ids        (filter-listing-tool-ids (set (keys perms)) public-tool-ids params)]
-    {:tools
+        tool-ids        (filter-listing-tool-ids (set (keys perms)) public-tool-ids params)
+        paged-params    (assoc params :tool-ids tool-ids :deprecated false)
+        unpaged-params  (dissoc paged-params :limit :offset)
+        total           (count (persistence/get-tool-listing unpaged-params))]
```
psarando: It's probably faster to have a db query perform the total count, rather than fetching everything and counting it in the code (see `apps.persistence.app-listing` for some examples); but that can be saved for a future PR.

We could add a `TODO` comment to this function for now, though, mentioning that we probably want to update this at some point to use a db query for the total count.

ash3rz: Cool, I was thinking there was probably a better way to go about this.  I think I might go with this as is so I can finish the UI work, then circle back. And now I'll know to check app-listing for reference 👍

ash3rz:
Thanks for the review!

PR #184, opened 2019-08-22
CORE-463 Merge /admin/tools function with /tools

ash3rz:


src/apps/tools.clj
```diff
@@ -69,10 +69,12 @@
...
   "Obtains a listing of tools accessible to the given user."
-  [{:keys [user] :as params}]
+  [{:keys [user] :as params} admin?]
   (let [public-tool-ids (perms-client/get-public-tool-ids)
         perms           (perms-client/load-tool-permissions user)
-        tool-ids        (filter-listing-tool-ids (set (keys perms)) public-tool-ids params)
+        tool-ids        (if admin?
+                          (admin-filter-listing-tool-ids public-tool-ids params)
+                          (filter-listing-tool-ids (set (keys perms)) public-tool-ids params))
         paged-params    (assoc params :tool-ids tool-ids :deprecated false)
```
psarando: This `list-tools` function is always filtering out `deprecated` tools (such as those that use the `backwards-compat` container). This prevents normal users from selecting a deprecated tool when creating a new app.

The `admin-list-tools` function did not set this `deprecated` flag to `false`, however, so admins can still list all tools.
The admin function also called `remove-nil-vals` on the `params` with the `tool-ids` added, since `admin-filter-listing-tool-ids` can potentially return a `nil` value.

ash3rz: Ah, good catches.  Will fix!

ash3rz:
Thanks for the review!

PR #188, opened 2019-09-05
CORE-468 Update POST /analyses to accept user requirement requests.

psarando:
This PR will update the `POST /analyses` endpoint to accept user `min` requirement requests (added by cyverse-de/common-swagger-api#47).

User `min` requirement requests submitted to `POST /analyses` will update that step's tool component `min` settings that are submitted in the job JSON sent in Condor submissions.

The `min` settings will be capped to the tool component's corresponding `max` setting.

For example, suppose a tool's container has resource requirements set in the db like the following:
```json
{
  "min_cpu_cores": 1,
  "max_cpu_cores": 4,
  "min_memory_limit": 1GB,
  "memory_limit": 8GB,
  "min_disk_space": 1GB,
  "step_number": 0
}
```
When a user submits a job, they can pick their resource requests as a value between the min and max (there is no per-tool max for disk, so this PR still needs to be updated to use a default max of `2TB` for now):
```json
{
  "min_cpu_cores": 2,
  "min_memory_limit": 4GB,
  "min_disk_space": 16GB,
  "step_number": 0
}
```
The `POST /analyses` endpoint will combine the user's selections with the tool's settings
for the job submission JSON:
```json
{
  "min_cpu_cores": 2,
  "max_cpu_cores": 4,
  "min_memory_limit": 4GB,
  "memory_limit": 8GB,
  "min_disk_space": 16GB,
  "step_number": 0
}
```
The `min` settings are intended to be used for condor `request_X` values for node matching, and the `max` settings will be used by `road-runner` for run-time container limits.


psarando:
Thanks for the review! 👍

PR #190, opened 2019-09-23
CORE-564: used lein-cljfmt to reformat everything

slr71:
I wanted to submit this as a PR just because of the sheer number of changes. Also, I'm spamming everyone just to make sure that everyone agrees on the formatting. I used the default settings for `lein-cljfmt`.

### DIFF
--- a/src/apps/metadata/tool_requests.clj
+++ b/src/apps/metadata/tool_requests.clj
@@ -172,8 +172,8 @@
   "Formats a single status record for a tool request."
   [req-status]
   (assoc req-status
-    :status_date (.getTime (:status_date req-status))
-    :comments    (or (:comments req-status) "")))
+         :status_date (.getTime (:status_date req-status))
+         :comments    (or (:comments req-status) "")))
 
 (defn- get-tool-request-details
   "Retrieves the details of a single tool request from the database."
@@ -228,7 +228,7 @@
 (defn- format-tool-request-dates
   [{:keys [date_submitted date_updated] :as tool-request}]
   (assoc tool-request :date_submitted (.getTime date_submitted)
-                      :date_updated   (.getTime date_updated)))
+         :date_updated   (.getTime date_updated)))
 
 (defn list-tool-requests
   "Lists tool requests."
--- a/src/apps/metadata/tool_requests.clj
+++ b/src/apps/metadata/tool_requests.clj
@@ -227,7 +227,8 @@
 
 (defn- format-tool-request-dates
   [{:keys [date_submitted date_updated] :as tool-request}]
-  (assoc tool-request :date_submitted (.getTime date_submitted)
+  (assoc tool-request
+         :date_submitted (.getTime date_submitted)
          :date_updated   (.getTime date_updated)))
 
 (defn list-tool-requests
--- a/src/apps/routes/analyses.clj
+++ b/src/apps/routes/analyses.clj
@@ -25,8 +25,8 @@
     :return listing-schema/AnalysisList
     :summary listing-schema/AnalysesListingSummary
     :description listing-schema/AnalysesListingDocs
-       ;; JSON query params are not currently supported by compojure-api,
-       ;; so we have to decode the String filter param and validate it here.
+    ;; JSON query params are not currently supported by compojure-api,
+    ;; so we have to decode the String filter param and validate it here.
     (ok (coerce! listing-schema/AnalysisList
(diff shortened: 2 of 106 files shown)


slr71:
I also added the one other change that I wanted to make for CORE-564 to this PR. That change was merely to switch to the latest version of `mescal`.

src/apps/metadata/tool_requests.clj
```diff
@@ -228,7 +228,7 @@
 (defn- format-tool-request-dates
   [{:keys [date_submitted date_updated] :as tool-request}]
   (assoc tool-request :date_submitted (.getTime date_submitted)
-                      :date_updated   (.getTime date_updated)))
+         :date_updated   (.getTime date_updated)))
```
psarando: I like that it will align k-v pairs under the map of an `(assoc map` line, but in this case it looks like the first k-v pair needs to be on a newline now, for consistency?

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -56,73 +55,73 @@
         job-config (substitute-param-values path-map (:job_config submission config))
         config     (substitute-param-values path-map config)]
     (assoc submission :job_config job-config
```
psarando: This `:job_config job-config` k-v pair could also be moved to a new line for consistency.

src/apps/service/apps/jobs/submissions/async.clj
```diff
@@ -56,73 +55,73 @@
...
-      (assoc results :parent-status parent-status
-                     :total         (inc total)
-                     :batch-status  (update batch-status job-status (fnil inc 0))))))
+   (let [parent-status (if (jp/completed? parent-status)
+                         parent-status
+                         (jp/get-job-status parent-id))
+         job-status    (if (jp/completed? parent-status)
+                         jp/canceled-status
+                         (submit-job-in-batch apps-client user submission total path-map))]
+     (assoc results :parent-status parent-status
```
psarando: New line for this k-v pair as well?

src/apps/tools/private.clj
```diff
@@ -31,8 +31,8 @@
                                               memory_limit (cfg/private-tool-memory-limit)}
          :as   container}]
   (assoc container :network_mode (if (= type "interactive") "bridge" "none")
```
psarando: Another new line for this k-v-pair?

src/apps/routes/analyses.clj
```diff
@@ -21,114 +21,114 @@
...
-       :query [{:keys [filter] :as params} AnalysisListingParams]
-       :return listing-schema/AnalysisList
-       :summary listing-schema/AnalysesListingSummary
-       :description listing-schema/AnalysesListingDocs
+    :query [{:keys [filter] :as params} AnalysisListingParams]
+    :return listing-schema/AnalysisList
+    :summary listing-schema/AnalysesListingSummary
+    :description listing-schema/AnalysesListingDocs
        ;; JSON query params are not currently supported by compojure-api,
        ;; so we have to decode the String filter param and validate it here.
```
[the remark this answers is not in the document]
slr71: Ah, yeah, I didn't notice that. I suspect it's configured not to alter comment indentation at all. We may be able to find a configuration setting for that, but we can probably do that later. 😁

src/apps/metadata/tool_requests.clj
slr71: Yeah, it's indenting `assoc` just like any other function call. Arguments on subsequent lines are aligned with the first argument on the first line, which happens to be the map itself in this case. Having the first k-v pair on a separate line seems reasonable. The only exception would be an assoc in a threading macro, I think. In that case, the newline can be optional.

slr71:
A git pre-commit hook is probably the most reliable way to go. I'm pretty sure I can get my editor to format code the same way, but I was also going to configure it to run `lein-cljfmt` whenever I save a file.

slr71:
Thanks for the reviews!

slr71:
As an aside, you can add a pre-commit hook by adding the following file to your local clone of the repository under the name `.git/hooks/pre-commit`:

```shell
#!/bin/sh

function print_format_error_msg() {
    cat <<EOF

Error: some code formatting issues were detected. Please fix them either
manually or by runinng the following command:

    lein cljfmt fix

EOF
    exit 1
}

lein cljfmt check || print_format_error_msg
```

PR #191, opened 2019-10-09
CORE-589: when processing job status updates, only retrieve the updat…

slr71:
…ed job step information when necessary

### DIFF
--- a/src/apps/service/apps.clj
+++ b/src/apps/service/apps.clj
@@ -244,11 +244,12 @@
               job         (jobs/lock-job job-id)
               batch       (when-let [parent-id (:parent_id job)] (jp/get-job-by-id parent-id))
               apps-client (get-apps-client-for-username (:username job))]
-          (doseq [{status :status sent-on :sent_on} updates]
-            (let [end-date (when (jp/completed? status) (str sent-on))
-                  job-step (first (jp/get-job-steps-by-external-id external-id))]
-              (when (jp/status-follows? status (:status job-step))
-                (jobs/update-job-status apps-client job-step job batch status end-date))))
+          (reduce (fn [job-step {status :status sent-on :sent_on}]
+                    (if (jp/status-follows? status (:status job-step))
+                      (let [end-date (when (jp/completed? status) (str sent-on))]
+                        (jobs/update-job-status apps-client job-step job batch status end-date))
+                      job-step))
+                  job-step updates)
           (jp/mark-job-status-updates-propagated (mapv :id updates))
           nil)))))
   ([job-id external-id status end-date]
--- a/src/apps/service/apps/de/jobs/common.clj
+++ b/src/apps/service/apps/de/jobs/common.clj
@@ -70,9 +70,9 @@
   [container requirements]
   (-> container
       (merge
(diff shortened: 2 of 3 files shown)


src/apps/service/apps.clj
```diff
@@ -244,11 +244,12 @@
...
-            (let [end-date (when (jp/completed? status) (str sent-on))
-                  job-step (first (jp/get-job-steps-by-external-id external-id))]
-              (when (jp/status-follows? status (:status job-step))
-                (jobs/update-job-status apps-client job-step job batch status end-date))))
+          (reduce (fn [job-step {status :status sent-on :sent_on}]
+                    (if (jp/status-follows? status (:status job-step))
+                      (let [end-date (when (jp/completed? status) (str sent-on))]
+                        (jobs/update-job-status apps-client job-step job batch status end-date))
+                      job-step))
+                  job-step updates)
```
slr71: This is the change that we discussed in Slack. One thing that I was concerned about was whether or not `jobs/lock-job-step` returns the same map format as `jp/get-job-steps-by-exteral-id` It appears that it does.

src/apps/service/apps/de/jobs/common.clj
```diff
@@ -70,9 +70,9 @@
   [container requirements]
   (-> container
       (merge
-        (merge-with max
-                    (filter-min-requirement-keys container)
-                    (filter-min-requirement-keys requirements)))
+       (merge-with max
+                   (filter-min-requirement-keys container)
+                   (filter-min-requirement-keys requirements)))
```
slr71: This is just a format change made by `lein cljfmt`.

src/apps/service/apps/jobs.clj
```diff
@@ -82,13 +82,14 @@
   (let [end-date (db/timestamp-from-str end-date)]
     (.updateJobStatus apps-client job-step job status end-date)
     (when batch (update-batch-status batch end-date))
-    (send-job-status-update apps-client (or batch job) job-step)))
+    (send-job-status-update apps-client (or batch job) job-step)
+    (first (jp/get-job-steps-by-external-id (:external_id job-step)))))
```
slr71: This function has been modified to return the updated job step information.

src/apps/service/apps/jobs.clj
```diff
@@ -82,13 +82,14 @@
...
 
 (defn update-job-status
   [apps-client job-step job batch status end-date]
   (cond
-    (jp/completed? (:status job))      (log-spurious-job-update job)
-    (jp/completed? (:status job-step)) (log-spurious-job-step-update job job-step)
+    (jp/completed? (:status job))      (do (log-spurious-job-update job) job-step)
+    (jp/completed? (:status job-step)) (do (log-spurious-job-step-update job job-step) job-step)
```
slr71: This function has also been modified to return the updated step information. We don't need to do another lookup in the first two cases, though.

slr71:
Thanks for the review!

slr71:
Thanks for the review! And I agree, there's a _lot_ of code in the apps service to unpack. 😄

PR #193, opened 2019-10-12
CORE-590: modify the job deletion endpoints to also cancel the analys…

slr71:
…is if it's not in a terminal status


slr71:
Thanks for the review!

PR #196, opened 2019-11-15
CORE-633 Add submitted resources reqs as defaults in relaunch response

psarando:
This PR will update the `GET /analyses/{analysis-id}/relaunch-info` endpoint to include `default_memory`, `default_cpu_cores`, and `default_disk_space` values in each step's resource `requirements` settings, if the corresponding `min` values were submitted in the original analysis.

Merging requires cyverse-de/common-swagger-api#53 first.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,7 @@
                  [org.cyverse/metadata-client "3.1.1"]
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/common-cfg "2.8.1"]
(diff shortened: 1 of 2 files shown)


psarando:
Also, I should point out that if we use these `default` keys in the future for app or tool resource defaults, these changes would mean that the `min` values requested in the original job would override any defaults loaded from the tool or the app in this relaunch response.

I think this would be the expected behavior.

psarando:
Thanks for the reviews! 👍

PR #197, opened 2019-12-20
CORE-661 Add `POST /analyses/relauncher` endpoint

psarando:
This PR will add the new `POST /analyses/relauncher` endpoint.

From the endpoint docs added in cyverse-de/common-swagger-api#54:

This service automatically relaunches analyses with the IDs given in the request.

* If a requested analysis was a sub-job of an HT analysis, then the relaunched analysis will also be a sub-job of that parent HT analysis.
* Any of the original analyses that have the `create_output_subdir` flag set to `false` will cause the corresponding relaunched analysis to use an output folder with `-redo-###` appended to the original output folder name, where `###` will be a `1` by default. If the original output folder name also ends with `-redo-###`, then the relaunched folder name will use a number 1 larger than the original. This will always apply to output folders of relaunched sub-jobs of an HT analysis.
  
  For example, if the output folder was `my-analysis`, then the relaunched output folder will use `my-analysis-redo-1`. If that relaunched analysis is in turn relaunched itself, then the next relaunched output folder will be `my-analysis-redo-2`.
* The names of relaunched HT analysis sub-jobs will also be renamed in this manner.

Also, in order to avoid loss of data, this endpoint will return an `EXISTS` error if any job "redo" output dirs already exist.

## Known Issues

If a user relaunches a job with the `create_output_subdir` flag set to `false` (such as in HT analyses sub-jobs), then relaunches that same job before the first relaunch has a chance to start running, then the output folder of the first relaunch won't have a chance to be created before the second relaunch is requested, the second relaunch will not return an `EXISTS` error, and both relaunched jobs will end up sharing the same output directory.

### DIFF
--- a/src/apps/service/apps/jobs/resubmit.clj
+++ b/src/apps/service/apps/jobs/resubmit.clj
@@ -1,10 +1,40 @@
 (ns apps.service.apps.jobs.resubmit
-  (:require [apps.service.apps.jobs.submissions :as submissions]
+  (:use [slingshot.slingshot :only [try+ throw+]])
+  (:require [apps.clients.data-info :as data-info]
+            [apps.service.apps.jobs.submissions :as submissions]
             [apps.service.apps.jobs.submissions.async :as async]
             [cheshire.core :as cheshire]
             [clojure.string :as string]
+            [clojure.tools.logging :as log]
             [clojure-commons.exception-util :as exception-util]
             [kameleon.uuids :as uuids]))
+
+(defn- get-paths-exist
+  [user paths]
+  (try+
+   (data-info/get-paths-exist user paths)
+   (catch Object _
+     (log/error (:throwable &throw-context)
+                "job resubmission failed: Could not lookup output folder existence.")
+     (throw+))))
+
+(defn- filter-existing-output-paths
+  [user output-paths]
+  (as-> output-paths p
+    (remove string/blank? p)
+    (get-paths-exist user p)
+    (:paths p)
+    (group-by val p)
+    (get p true)
+    (map first p)))
+
+(defn- find-existing-output-dirs
+  [user jobs]
+  (->> jobs
+       (map :submission)
+       (remove #(:create_output_subdir % true))
+       (map :output_dir)
+       (filter-existing-output-paths user)))
 
 (defn- decode-job-submission
   [{:keys [id submission] :as job}]
@@ -45,7 +75,11 @@
 (defn resubmit-jobs
   [apps-client user jobs]
   (let [jobs                  (map (comp format-resubmission decode-job-submission) jobs)
-        [ht-jobs single-jobs] ((juxt filter remove) :is_batch jobs)]
+        [ht-jobs single-jobs] ((juxt filter remove) :is_batch jobs)
+        existing-output-dirs  (find-existing-output-dirs user jobs)]
+    (when-not (empty? existing-output-dirs)
+      (exception-util/exists "One or more output folders already exist." :paths existing-output-dirs))
+
     (when-not (empty? single-jobs)
       (async/resubmit-jobs apps-client user single-jobs))
     (doseq [ht-job ht-jobs]
--- a/src/apps/service/apps/jobs/resubmit.clj
+++ b/src/apps/service/apps/jobs/resubmit.clj
@@ -1,0 +1,52 @@
+(ns apps.service.apps.jobs.resubmit
+  (:require [apps.service.apps.jobs.submissions :as submissions]
+            [apps.service.apps.jobs.submissions.async :as async]
+            [cheshire.core :as cheshire]
+            [clojure.string :as string]
+            [clojure-commons.exception-util :as exception-util]
+            [kameleon.uuids :as uuids]))
+
+(defn- decode-job-submission
+  [{:keys [id submission] :as job}]
+  (when-not submission (exception-util/not-found (str "Job submission values could not be found for " id)))
+  (update job :submission #(-> % .getValue (cheshire/decode true))))
+
+(defn- format-redo-name
+  "Returns a string by appending `-redo-1` to `original-name`.
+   If `original-name` already ends with `-redo-###` (where `###` is any digit),
+   then returns `original-name-redo-###` where `###` is a number 1 larger than the original trailing digit."
+  [original-name]
+  (let [regex          #"(.*-redo-)(\d+)$"
+        format-matches #(str %1 (-> %2 Integer/parseInt inc))]
+    (if (re-matches regex original-name)
+      (string/replace original-name regex #(format-matches (%1 1) (%1 2)))
+      (str original-name "-redo-1"))))
+
+(defn- format-resubmission-output-dir
+  [{:keys [submission] :as job}]
+  (if (:create_output_subdir submission true)
+    job
+    (update-in job [:submission :output_dir] format-redo-name)))
+
+(defn- format-resubmission-name
+  [{:keys [parent_id] :as job}]
+  (if parent_id
+    (-> job
+        (update-in [:submission :parent_id] uuids/uuidify)
+        (update-in [:submission :name] format-redo-name))
+    job))
+
+(defn- format-resubmission
+  [job]
+  (-> job
+      format-resubmission-output-dir
+      format-resubmission-name))
+
+(defn resubmit-jobs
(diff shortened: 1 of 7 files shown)


src/apps/service/apps/jobs/resubmit.clj
```diff
@@ -0,0 +1,86 @@
...
+    (:paths p)
+    (group-by val p)
+    (get p true)
+    (map first p)))
+
+(defn- find-existing-output-dirs
+  [user jobs]
+  (->> jobs
+       (map :submission)
+       (filter #(not (:create_output_subdir % true)))
```
slr71: Minor suggestion: you could also use `(remove (:create_output_subdir % true))`

src/apps/service/apps/jobs/resubmit.clj
```diff
@@ -0,0 +1,86 @@
...
+
+(defn- filter-existing-output-paths
+  [user output-paths]
+  (as-> output-paths p
+    (remove string/blank? p)
+    (get-paths-exist user p)
+    (:paths p)
+    (group-by val p)
+    (get p true)
+    (map first p)))
```
slr71: 👍

src/apps/service/apps/jobs/resubmit.clj
psarando: Of course! That's cleaner, thanks for the suggestion ✨

slr71: Oops! I forgot a hash mark in my example. Is should be `(remove #(:create_output_subdir % true))`. Listen to what I mean, not what I type. 😆

src/apps/service/apps/jobs/resubmit.clj
```diff
@@ -0,0 +1,86 @@
...
+
+(defn- format-redo-name
+  "Returns a string by appending `-redo-1` to `original-name`.
+   If `original-name` already ends with `-redo-###` (where `###` is any digit),
+   then returns `original-name-redo-###` where `###` is a number 1 larger than the original trailing digit."
+  [original-name]
+  (let [regex          #"(.*-redo-)(\d+)$"
+        format-matches #(str %1 (-> %2 Integer/parseInt inc))]
+    (if (re-matches regex original-name)
+      (string/replace original-name regex #(format-matches (%1 1) (%1 2)))
```
ash3rz: I'm struggling with `#(format-matches (%1 1) (%1 2))`.  I understand the regex should give you 2 matching groups and we want to send both groups to `format-matches`, but I'm not seeing how `(%1 1) (%1 2)` translates in clojure-fu to that.  I would think it would be `#(format-matches $1 $2)` or something similar

src/apps/service/apps/jobs/resubmit.clj
```diff
@@ -0,0 +1,86 @@
...
+(defn- format-resubmission
+  [job]
+  (-> job
+      format-resubmission-output-dir
+      format-resubmission-name))
+
+(defn resubmit-jobs
+  [apps-client user jobs]
+  (let [jobs                  (map (comp format-resubmission decode-job-submission) jobs)
+        [ht-jobs single-jobs] ((juxt filter remove) :is_batch jobs)
```
ash3rz: `juxt` and `comp` - neat! 
Probably won't remember these myself when I could use them, but very cool 😄

src/apps/service/apps/jobs/resubmit.clj
ianmcorvidae: So! This is where `macroexpand` can be useful in learning how this sort of thing works.

From a `lein repl` in the root directory of the apps repo:

```
apps.core=> (macroexpand `(string/replace original-name regex #(format-matches (%1 1) (%1 2))))
(string/replace apps.core/original-name apps.core/regex (fn* [p1__4106__4107__auto__] (apps.core/format-matches (p1__4106__4107__auto__ 1) (p1__4106__4107__auto__ 2))))
```

This is a bit dense, so let's focus on the lambda by itself and format a bit:

```clojure
(fn* [p1__4106__4107__auto__]
  (apps.core/format-matches
    (p1__4106__4107__auto__ 1)
    (p1__4106__4107__auto__ 2)))
```

For good measure, let's rename the automatically-generated parameter names, because those are annoying. We'll call it `match-map` because, well, it's a map of matches from the regex, I think. It might be an array, not sure, since the indices are integers.

```clojure
(fn* [match-map]
  (apps.core/format-matches
    (match-map 1)
    (match-map 2)))
```

You can ignore the whole `fn*` thing, that's a clojure internal and for our purposes here it's equivalent to plain `fn` for an anonymous function.

Now, when maps are put in the function position in clojure, the first argument is the index into the map to pull out. So what this does is matches against the regex and passes the first group as the first argument and the second group as the second argument to `format-matches`.

I see that you're looking at `$1` and `$2` and that's a bit of a confusing thing. You can use those things if you're issuing a _string_ as a replacement, but not if you're passing a function -- they're a thing provided by the regex stuff (via Java) and not by Clojure. The reason you can't use it here is that the second group needs to be parsed as an integer, then incremented, before putting it back into the string, hence the `(-> %2 Integer/parseInt inc)` in `format-matches`. So there isn't the option of sticking the exact thing matched right back into the replacement, because it needs to be transformed first. Hence the function.

Hopefully that's informative, lol, I was mostly wanting to show off `macroexpand` a little!

ianmcorvidae: maybe the tl;dr here is just that the $ variables are only available when replacing with a string, you have to fetch the matched groups out of the match object manually when you're using a function. But hopefully the macroexpand madness will help show why.

ianmcorvidae: as another clojure detail, `#()` is a _reader_ macro which is a little weirder, but it does just end up kinda working the same. see https://clojure.org/reference/reader#_dispatch for this bit, though admittedly I don't myself know how exactly that translation happens. macroexpand will still expand those macros, though not actually because that's its job.

psarando: Thanks for the tip on `macroexpand` @ianmcorvidae! I keep forgetting about it 🤦‍♂ 

Just to append some practical examples to that already thorough explanation:

The regex matcher will return a list. For example:
```clojure
apps.core=> (re-matches #"(.*-redo-)(\d+)$" "hello-world-redo-1")
["hello-world-redo-1" "hello-world-redo-" "1"]
```
If that list was named `matches`, then `(matches 1)` would return "hello-world-redo-" and `(matches 2)` would return "1".

The `#(format-matches (%1 1) (%1 2))` anonymous function would be the same as the following:
```clojure
(fn [matches] (format-matches (matches 1) (matches 2)))
```
So when that list is passed to that function, it's the same as calling
```clojure
(format-matches "hello-world-redo-" "1")
```
Now, `format-matches` is also defined as an anonymous function, `#(str %1 (-> %2 Integer/parseInt inc))`, which would be the same as something like the following:
```clojure
(fn [folder-redo-name redo-suffix] (str folder-redo-name (-> redo-suffix Integer/parseInt inc)))
```
So in this example, that's the same as calling
```clojure
(str "hello-world-redo-" (-> "1" Integer/parseInt inc))
```
Which returns "hello-world-redo-2" :qed:

ash3rz: Thanks for all the really detailed explanations!  As expected, with a good night's rest and fresh eyes, I see the light 😄 I think I was just focusing too much on the regex and neglecting to really parse the anonymous function aspect so `(%1 1) (%1 2)` was not clicking for me, but now it makes sense.

`macroexpand` seems neat!  I think it takes a little bit of getting used to, like how Ian had to reformat it to look a little less hairy, but I can see the utility in it 👍 I'll have to remember to test that out for myself, like I wrote in my notes lol

Thanks again all!

psarando:
Notifications for normal relaunched jobs are coming through as expected, but when relaunching subjobs of an HT analysis (the main purpose of this new feature), a batch/bundled notification is not being sent when all relaunched subjobs are complete (as is done when all subjobs of the original HT analysis are complete).

I'll leave this PR open until I investigate re-sending the bundled notification for the parent HT analysis.

psarando:
It looks like just resetting the parent HT analysis back to the `Submitted` state is enough to allow the new subjob state changes to update the parent once again.

psarando:
Thanks for the review! 👍

PR #198, opened 2020-04-07
CORE-971: added an endpoint to list a single app in the DE

slr71:
The changes in this PR add a new endpoint that can be used to obtain app listing information for a single app.

The purpose of adding a separate endpoint for listing a single app in the DE is also to add clarity. Adapting the existing app search endpoint to work for listing a specified app ID seems potentially confusing. For example, what happens if the app-id and search parameters are both specified and they conflict (that is the app ID matches an app, but the search string would exclude the app)? I can imagine a valid argument for excluding the app because it doesn't match the search string. After all, maybe the caller only wants to list that specific app if it also matches the search string. I can also imagine a valid argument for including the app because the app-id parameter should override the search string. Given the confusion inherent in interactions between the app-id and search parameters (or the lack thereof), I decided it would probably be best to create a separate endpoint.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,7 @@
                  [org.cyverse/metadata-client "3.1.1"]
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/common-cfg "2.8.1"]
-                 [org.cyverse/common-swagger-api "3.0.1-SNAPSHOT"]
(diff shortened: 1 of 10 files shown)


sri2k1us:
LGTM..Thanks for getting it done 👍

slr71:
Thanks for the reviews!

PR #199, opened 2020-05-19
Parallelize creation of virtual categories

ianmcorvidae:
also, remove non-public categories faster with futures and doall

fairly small PR, but improves performance of the main categories endpoint substantially. apps was doing some significant work it didn't need to, plus this parallelizes some of what's left. Here's what's going on:

`count-deleted-and-orphaned-apps` is pretty slow (about 5s from here, though it's probably at least a little better not over VPN). But, it's only used for the Trash category, which has `:is_public` set to true. `format-private-virtual-groups` removes things with `:is_public` true, but only after they've been calculated out, so we were just calculating that all out to immediately toss it. Instead, I've changed it to use a future for all the counts, and then deref the futures after the `remove` call. Also, I added in a `doall` so it would start the whole process before leaving `format-private-virtual-groups` (before, it was calculating it much later, during response coercion, due to the lazy sequences).

I think that it'll still calculate out the future with `count-deleted-and-orphaned-apps`, but now it does it in another thread and never derefs it, which at least means the endpoint returns without it needing to have calculated that count.

### DIFF
--- a/src/apps/service/apps.clj
+++ b/src/apps/service/apps.clj
@@ -8,6 +8,7 @@
         [service-logging.thread-context :only [with-logging-context]])
   (:require [clojure.tools.logging :as log]
             [clojure-commons.exception-util :as cxu]
+            [clojure.walk :as walk]
             [mescal.de :as agave]
             [apps.clients.notifications :as cn]
             [apps.persistence.jobs :as jp]
@@ -76,7 +77,9 @@
 (defn get-app-categories
   [user params]
   (let [client (get-apps-client user "type=apps")]
-    {:categories (transaction (.listAppCategories client params))}))
+    {:categories (transaction
+                   (walk/prewalk (fn [x] (if (or (future? x) (delay? x)) (deref x) x))
+                     (.listAppCategories client params)))}))
 
 (defn list-apps-in-category
   [user system-id category-id params]
--- a/src/apps/service/apps.clj
+++ b/src/apps/service/apps.clj
@@ -364,7 +364,8 @@
 
 (defn get-admin-app-categories
   [user params]
-  {:categories (.getAdminAppCategories (get-apps-client user) params)})
+  (walk/prewalk (fn [x] (if (or (future? x) (delay? x)) (deref x) x))
+    {:categories (.getAdminAppCategories (get-apps-client user) params)}))
 
 (defn search-admin-app-categories
   [user params]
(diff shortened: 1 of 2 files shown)


src/apps/service/apps.clj
```diff
@@ -76,7 +77,9 @@
 (defn get-app-categories
   [user params]
   (let [client (get-apps-client user "type=apps")]
-    {:categories (transaction (.listAppCategories client params))}))
+    {:categories (transaction
+                   (walk/prewalk (fn [x] (if (or (future? x) (delay? x)) (deref x) x))
```
ianmcorvidae, line 81: This makes it so any futures or delays are deref'd before the transaction ends. This is to ensure that any queries therein have started before the transaction ends, and also to replace their values with the non-delay/non-future version. Basically, force everything to have run before ending the transaction.

src/apps/service/apps/de/listings.clj
```diff
@@ -145,19 +145,24 @@
...
   [user group workspace params]
   (let [virtual-groups (format-private-virtual-groups user workspace params)
-        actual-count   (count-apps-in-group-for-user
-                        (:id group)
-                        (:username user)
-                        params)]
+        actual-count   (future (count-apps-in-group-for-user
+                                 (:id group)
+                                 (:username user)
+                                 params))]
```
ianmcorvidae, line 165: I realized this could be made a future too, because nothing depends on its value until it's returned to the client.

src/apps/service/apps/de/listings.clj
```diff
@@ -97,7 +97,7 @@
    :id         trash-category-id
    :name       "Trash"
    :is_public  true
-   :total      (count-deleted-and-orphaned-apps params)})
+   :total      (delay (count-deleted-and-orphaned-apps params))})
```
ianmcorvidae, line 100: This is a delay now, because otherwise it always executes and it takes ages, and can't get cancelled due to the transaction, it seems. Really, I think the trash category stuff only happens when the category is listed directly, so it won't affect much.

src/apps/service/apps/de/listings.clj
ianmcorvidae: (to be clear here, the prewalks ensure this is made into a real value before returning to the client. Assuming I didn't miss any, anyway)

ianmcorvidae:
Now that I've fixed things that were wrong originally, I'll go ahead and merge this. Thanks for the review, even if I was a bit premature in it D:

PR #200, opened 2020-06-12
WIP CORE-600 Add total count of jobs by status

ash3rz:
Wanted some eyes on this before I try to make other changes.  This produces something like:
```
{
  "analyses": [ ... ],
  "timestamp": "1591988105298",
  "status-count": [
    {
      "count": 23,
      "status": "Submitted"
    },
    {
      "count": 3,
      "status": "Canceled"
    },
    {
      "count": 4,
      "status": "Completed"
    }
  ],
  "total": 30
}
```

### DIFF
--- a/src/apps/persistence/jobs.clj
+++ b/src/apps/persistence/jobs.clj
@@ -289,14 +289,28 @@
       (aggregate (count :*) :count)
       (add-internal-app-clause include-hidden)))
 
+(defn- count-jobs-of-types-query
+  [username filter include-hidden types accessible-ids]
+  (-> (select* (add-job-query-filter-clause (count-jobs-base include-hidden) username filter))
+      (where {:j.deleted false})
+      (where {:j.id (util/sqlfn-any-array "uuid" accessible-ids)})
+      (where (not (exists (job-type-subselect types))))))
+
 (defn count-jobs-of-types
   "Counts the number of undeleted jobs of the given types in the database for a user."
   [username filter include-hidden types accessible-ids]
-  ((comp :count first)
-   (select (add-job-query-filter-clause (count-jobs-base include-hidden) username filter)
-           (where {:j.deleted false})
-           (where {:j.id (util/sqlfn-any-array "uuid" accessible-ids)})
-           (where (not (exists (job-type-subselect types)))))))
+  (-> (count-jobs-of-types-query username filter include-hidden types accessible-ids)
+      (select)
+      (first)
+      (:count)))
+
+(defn count-jobs-of-statuses
+  "Counts the number of undeleted jobs of the given types grouped by statuses in the database for a user."
+  [username filter include-hidden types accessible-ids]
+  (-> (count-jobs-of-types-query username filter include-hidden types accessible-ids)
+      (fields :j.status)
+      (group :j.status)
+      (select)))
 
 (defn- translate-sort-field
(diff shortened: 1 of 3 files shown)


ianmcorvidae:
Giving you a quick review, basic setup looks reasonable to me. Since the function now runs two queries that don't depend directly on each other, I'd recommend putting one or both in a `future` and then `deref`-ing it when needed (`@` is a clojure reader-macro for `deref` so you may have seen that). This would let the queries run in parallel. Small thing though, ultimately, I just have performance on the brain!

src/apps/persistence/jobs.clj
```diff
@@ -289,14 +289,28 @@
       (aggregate (count :*) :count)
       (add-internal-app-clause include-hidden)))
 
+(defn- count-jobs-of-types-query
+  [username filter include-hidden types accessible-ids]
+  (-> (select* (add-job-query-filter-clause (count-jobs-base include-hidden) username filter))
```
ianmcorvidae, line 294: Oh, one other little thing. Since this includes the `filter` param as passed in (and the type filters, etc.), it'll show the counts for whatever's visible, and not for everything by the user. I'm not sure from the ticket which is intended -- are we showing counts-by-status for the current filtered view, or count-by-status for the user as a whole, regardless of the specific filter on the analyses window/analyses view?

ash3rz: That's a really good question.  Some context is that this is also hopefully going to help with CORE-637.  In which case, knowing how many condor jobs vs vice jobs would be necessary.  You could get that information with this currently and applying a filter, or perhaps it would be better to have the counts grouped by type and then status, shown for all jobs, and not factor in the filter at all?

ash3rz: (Not the notification aspect of that ticket, if we also want that, but being able to display those values in the Analyses window)

ianmcorvidae: Yeah -- 637 suggests that it's supposed to be for everything, not just whatever the view is filtered down to, and for that it'd probably be nice to group by both type and status. But I can somewhat see potential value in seeing a summary of the current view, too. I'm not sure which is more consistent with other endpoints, or if any of this suggests these counts should go in a separate endpoint or anything like that.

(which means I'm annoyingly asking questions I can't answer, but I guess that's what review is for sometimes, heh)

ash3rz: Ooh, now that you mention it, I'm remembering that Nirav wanted totals like this in the dashboard as well.  Maybe that suggests it should be its own endpoint that `dashboard-aggregator` can hit?

ianmcorvidae: Hm, good question. `dashboard-aggregator` of course already lists some analysis information, but it appears to do it by querying the DB itself, not by proxying anything to the apps service. So I don't know that it'd use such an endpoint anyway (though maybe it should? unclear to me)

ash3rz: OK, now I'm reading back through my convo with Sriram and I focused a little too much on just what CORE-600 described.  There was also mention of including a time period parameter and displaying this data for all of the DE for logged out users in the dashboard.  So, I have a start, but this will require more :smile:  I feel now that kind of warrants a separate endpoint.

ash3rz:
Thanks for the initial review! I will look into the performance improvement you mentioned and also look into making this its own endpoint based on discussions in Slack

PR #201, opened 2020-06-18
CORE-600 Add job status counts as an endpoint and to job listings

ash3rz:
In case we did want to switch to having the totals be grouped by job type (so, for example, an interactive section with all those status totals, and then an agave section with all those status totals, etc.), instead of showing the totals for whatever type filter is set, I've been looking into how to do that as well (if only for my own practice). It's taking me longer than I anticipated to figure that out, so I'm putting this up now in case we want to stick with this format anyway. So the end result is still looking like:
```
{
  "analyses": [ ... ],
  "timestamp": "1591988105298",
  "status-count": [
    {
      "count": 23,
      "status": "Submitted"
    },
    {
      "count": 3,
      "status": "Canceled"
    },
    {
      "count": 4,
      "status": "Completed"
    }
  ],
  "total": 30
}
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,7 @@
                  [org.cyverse/metadata-client "3.1.1"]
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/common-cfg "2.8.1"]
-                 [org.cyverse/common-swagger-api "3.0.1"]
+                 [org.cyverse/common-swagger-api "3.0.3-SNAPSHOT"]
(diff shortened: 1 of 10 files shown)


ash3rz:
Cool, thanks for the review!  I haven't heard anybody having a preference either way, so I think I'll just continue with this.  I should have a PR with the accompanying `terrain` and `common-swagger-api` changes soon.

ash3rz:
Thanks for the reviews!

PR #202, opened 2020-06-26
temporary changes to speed up admin app listings

slr71:


slr71:
Thanks for the review!

PR #203, opened 2020-07-15
CORE-1094: made the IP address optional for the login endpoint

slr71:


slr71:
Thanks for the review! I'm going to push one more commit to switch to the latest release version of `common-swagger-api` before merging this PR.

PR #204, opened 2020-07-20
CORE-1102: perform DE an HPC analysis listings concurrently.

slr71:
I tested these changes locally in the REPL using these functions:

``` clojure
(defn test-search
  [term]
  (binding [current-user user]
    (apps/search-apps user {:search term})
    nil))

(defn test-admin-search
  [term]
  (binding [current-user user]
    (apps/admin-search-apps user {:search term})
    nil))

(def ontology-version "http://edamontology.org/1.16/2016/12/19/15-54-46.2")
(def root-iri "http://edamontology.org/operation_2928")
(def attr "rdf:type")
(def sort-field "name")
(def sort-dir "ASC")

(defn test-list-apps-under-hierarchy
  []
  (binding [current-user user]
    (apps/list-apps-under-hierarchy
     user
     root-iri
     attr
     {:type       attr
      :sort-field sort-field
      :sort-dir   sort-dir})
    nil))

(defn test-admin-list-apps-under-hierarchy
  []
  (binding [current-user user]
    (apps/admin-list-apps-under-hierarchy
     user
     ontology-version
     root-iri
     attr
     {:type       attr
      :sort-field sort-field
      :sort-dir   sort-dir})
    nil))

(def community-id "iplant:de:qa:communities:Metalheads")

(defn test-list-apps-in-community
  []
  (binding [current-user user]
    (apps/list-apps-in-community
     user
     community-id
     {:sort-field sort-field
      :sort-dir   sort-dir})
    nil))

(defn test-admin-list-apps-in-community
  []
  (binding [current-user user]
    (apps/admin-list-apps-in-community
     user
     community-id
     {:sort-field sort-field
      :sort-dir   sort-dir})
    nil))
```

I removed the `nil` return value to do some cursory checks on the results that were being returned.

### DIFF
--- a/src/apps/service/apps/combined/util.clj
+++ b/src/apps/service/apps/combined/util.clj
@@ -22,6 +22,12 @@
         (apply-offset params)
         (apply-limit params))))
 
+(defn combine-concurrent-app-listings
+  "Expects results to be a list of futures that can be dereferenced to return a map in a format like
+   {:total int, :apps []}"
+  [params results]
+  (combine-app-listings params (mapv deref results)))
+
 (defn get-apps-client
   ([clients]
    (or (first (filter #(.supportsSystemId % jp/de-client-name) clients))
--- a/conf/test/logback.xml
+++ b/conf/test/logback.xml
@@ -14,6 +14,9 @@
   <!-- Jargon Logging -->
   <logger name="org.irods.jargon" level="ERROR" />
 
+  <!-- Change level to DEBUG to log database queries. -->
+  <logger name="kameleon.util" level="WARN"/>
+
   <logger name="apps.service.callbacks" level="INFO" />
   <logger name="apps.util.json" level="DEBUG" />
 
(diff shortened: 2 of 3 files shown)


slr71:
Thanks for the review!

slr71:
Thanks for the reviews!

src/apps/service/apps/combined/util.clj
```diff
@@ -22,6 +22,12 @@
         (apply-offset params)
         (apply-limit params))))
 
+(defn combine-concurrent-app-listings
```
ianmcorvidae, line 25: I am, of course, late to the party, but I suppose it's maybe still worth noting: I don't know if it's worth it, but I'd have probably rolled this into `combine-app-listings` something like:

```clojure
(let [maybe-deref (fn [r] (if (future? r) (deref r) r))
       results (mapv maybe-deref results)
...etc
```

not sure if that's better or worse, but I do somewhat like a model where we expect things that take deref-able arguments to also accept already-deref'd values in the same function, such that it's mostly transparent.

(just food for thought, anyway!)

slr71: I like that idea. I'll add it in the next PR.

PR #205, opened 2020-07-21
CORE-1102: fixed a null pointer exception

slr71:


slr71:
Thanks for the review!

PR #207, opened 2020-07-29
CORE-1106: switched the endpoints for listing apps under a hierarchy …

slr71:
…to use the new queries

This change updates the app listing by ID endpoints to use the new queries. There isn't much of a performance improvement in this case, but I would eventually like to update all app listing code to use the same queries just to avoid potential confusion. There will be more PRs related to this coming up, but I wanted to get this one submitted because I'm going to be adding a couple of new features to the apps service over the next few days. I'm hoping to return to this task after I finish adding the new features.


slr71:
Heh, looks like I don't need to implement that new feature after all. I'm going to close this PR in order to continue work on this task.

PR #208, opened 2020-07-30
CORE-1106: more app listing improvements

slr71:
After doing some more testing, I think that this PR and the previous one combined make all of the changes that will be useful to make at this time. We could migrate other endpoints to the new listing queries, but that's going to require more effort than it's worth at the moment.

This PR updates the endpoints to list apps under an ontology hierarchy, and it removes the `:new` parameter that I had been using to switch between the old and new queries.


slr71:
Thanks for the review!

PR #209, opened 2020-07-31
CORE-1110 analysis listing performance improvements.

slr71:
This change reduced the number of queries that are made during analysis listing in order to improve the responsiveness of the listing endpoints.


slr71:
I suspected the reason would become apparent. 😆 Thanks for the review!

PR #211, opened 2020-10-30
CORE-1164: added the `type` and `path` members to parameters sent in …

slr71:
…outgoing job submissions

The purpose of this PR is to facilitate jobs using the iRODS CSI driver. Adding the full path to each input file will allow us to bind a directory to a FUSE mount pointing to `/iplant` or `/iplant/home` into the container and adjust the input file or folder paths accordingly. In the case of a pipeline, the paths don't need to be modified (because they refer to paths relative to the current working directory). Because of this, relative paths should not be adjusted when building the command line.

The `type` parameter currently isn't used, but I added it in case it became necessary in the future. (There was a period of time when I thought it was going to be necessary for this change, but we found an alternate solution.)

### DIFF
--- a/src/apps/service/apps/de/jobs/params.clj
+++ b/src/apps/service/apps/de/jobs/params.clj
@@ -119,7 +119,8 @@
      {:id    (:id param)
       :name  opt-arg
       :order (:order param 0)
-      :value (stringify opt-val)}))
+      :value (stringify opt-val)
+      :type  (:type param)}))
   ([param param-value]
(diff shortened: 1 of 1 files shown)


ash3rz (commented):
Seems straightforward 👍 
(Not at all familiar with job submission stuff, so leaving as just a comment)

psarando (approved):
LGTM :shipit:

slr71:
Thanks for the reviews!

PR #212, opened 2021-01-21
CORE-1216: update analysis listings in order to deprecate the `app_di…

slr71:
…sabled` field

Obtaining the value of the `app_disabled` field for analysis that run in Tapis requires users to authenticate to Tapis. This produces some inconsistent behavior because authentication is required only if one of the analyses in the current page of the listing happens to be a Tapis analysis. We're deprecating this field and updating the endpoints to always return `false` in this field.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -29,7 +29,7 @@
                  [org.cyverse/metadata-client "3.1.1"]
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/common-cfg "2.8.1"]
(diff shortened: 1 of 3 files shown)


ianmcorvidae (approved):
Looks good to me. Do we already check for disabled apps when relaunching, or is that another change that's coming after this PR?

slr71:
I think it's already present, but I have some other changes to make for app listings. I figured I'd check it when I'm looking at the other stuff.

slr71:
Thanks for the review!

PR #214, opened 2021-02-03
CORE-1221: added limit checks to the app listing and launch information retrieval endpoints.

slr71:


src/apps/persistence/app_listing.clj
```diff
@@ -3,13 +3,22 @@
...
         [kameleon.util :only [query-spy]]
         [kameleon.util.search]
         [korma.core :exclude [update]])
   (:require [apps.constants :as c]
-            [clojure.string :as str]
-            [apps.persistence.util :as util]))
+            [clojure.string :as str]))
+
+;; The `app_listing` view has an array aggregate column that we need to be able to extract. This tells
+;; `clojure.java.jdbc` how to convert the array to a (possibly nested) vector.
```
ianmcorvidae, line 15: :+1:

src/apps/service/apps/de/limits.clj
```diff
@@ -0,0 +1,42 @@
...
+            [apps.persistence.job-limits :as job-limits]))
+
+(def ^:private err-permission-needed "ERR_PERMISSION_NEEDED")
+(def ^:private err-forbidden "ERR_FORBIDDEN")
+(def ^:private err-limit-reached "ERR_LIMIT_REACHED")
+
+(defn- load-concurrent-vice-analysis-limit-check-results
+  "Loads the limit check for the number of concurrently running VICE analyses."
+  [user]
+  (let [limit-info (analyses/get-concurrent-job-limit (:shortUsername user))
```
ianmcorvidae, line 12: I know this is extremely minor, but for the sake of performance, it would be nice to make this HTTP call asynchronous by wrapping it in a `future` and then `deref` when it's used below, since the next line doesn't depend on this result and could be done in parallel.

src/apps/service/apps/de/listings.clj
```diff
@@ -297,27 +298,32 @@
...
 (defn- app-ids->beta-ids-set
   "Filters the given list of app-ids into a set containing the ids of apps marked as `beta`"
   [username app-ids]
   (let [beta-avu {:attr (workspace-metadata-beta-attr-iri)
                   :value (workspace-metadata-beta-value)}]
     (set (metadata-client/filter-by-avus username app-ids [beta-avu]))))
 
+(defn- get-app-listing-formatter
+  "Returns a function that can be used to format the listing for a single app. Using a higher order function for this
+   makes it easier to consolidate repetitive tasks without degrating performance."
```
ianmcorvidae, line 310: Could you explain this more? I don't have any problem with the change, but I wasn't sure why this approach was better than using `partial` like before, so I'm wondering what I'm not understanding.

src/apps/util/db.clj
```diff
@@ -0,0 +1,57 @@
...
+  "This namespace contains some database utility functions. For the time being, these functions will use Korma to
+   obtain database connections. This code will have to be modified when Korma is removed."
+  (:require [apps.util.conversions :refer [date->timestamp]]
+            [korma.core :as sql]
+            [korma.db :as db]))
+
+(defmacro transaction
+  "Executes queries within the body in a single transaction. Nested transactions are absorbed into the outermost
+   transaction. For now, this is just an alias to korma.db/transaction. When Korma is removed, it will be necessary
+   to come up with an alternative implementation."
```
ianmcorvidae, line 11: This is of course out of scope for this PR, but it might be nice to eventually do this with https://github.com/seancorfield/next-jdbc once Korma is out of the picture.

ianmcorvidae (approved):
A few random thoughts, one question, one super trivial performance nitpick. Basically, looks good to me :laughing:

johnworth (approved):
LGTM :shipit:

src/apps/service/apps/de/limits.clj
slr71: Good idea! I'll change that.

src/apps/service/apps/de/listings.clj
slr71: If I continued using `partial` like we were before, I would have had to add yet another parameter to the formatting function, which already has a lot of positional parameters. I thought briefly about changing the positional parameters to named parameters (and putting all of the named parameters into an option map so that the app listing itself is still a positional parameter). Ultimately, however, I decided to go with this approach because I also would have had to add code to obtain the limit checks to every place where the formatting function is being called. Obtaining the limit checks and the list of beta app IDs here, allows me to remove the (minor) amount of duplicate code from all of the app listing functions. There are trade-offs, of course. The returned anonymous function is another level of indirection to think about. I still think this helps to make the code a little bit easier to read, though.

ianmcorvidae: Makes sense. I think my mind got caught up on the mention of performance -- this and using `partial` are probably similar performance-wise, but this is definitely easier to read, and more performant than other ways of making it more easily readable, so I think I've connected the dots. Thanks!

slr71: Ah, yeah, I guess I didn't really need to mention performance. All I really meant was that we'd still be retrieving the beta ID list and the limit checks once even though those things aren't being done in the calling functions explicitly.

slr71:
Thanks for the reviews!

psarando (commented):
Since this is already merged, I didn't review this as thoroughly as I would have, but from what I understand, this LGTM 👍

PR #215, opened 2021-02-05
CORE-1221: removed a now unnecessary call to org.postgresql.jdbc.PgAr…

slr71:
…ray.getArray

There was a lingering call to `org.postgresql.jdbc.PgArray.getArray` that I missed after I added the `extend-protocol` statement in `apps.persistence.app-listings`, and it was causing an error.

### DIFF
--- a/src/apps/service/apps/job_listings.clj
+++ b/src/apps/service/apps/job_listings.clj
@@ -76,7 +76,7 @@
   [rep-steps job]
   (remove-nil-vals
    (assoc (format-base-job job)
-          :external_ids     (vec (.getArray (:external_ids job)))
+          :external_ids     (:external_ids job)
           :interactive_urls (interactive-urls job rep-steps))))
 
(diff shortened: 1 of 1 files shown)


ianmcorvidae (approved):
:shipit:

src/apps/service/apps/job_listings.clj
```diff
@@ -76,7 +76,7 @@
   [rep-steps job]
   (remove-nil-vals
    (assoc (format-base-job job)
-          :external_ids     (vec (.getArray (:external_ids job)))
```
slr71, line 79: The protocol extension defined in [apps.persistence.app-listings](https://github.com/cyverse-de/apps/blob/master/src/apps/persistence/app_listing.clj#L14-L21) tells `clojure.java.jdbc` how to convert an instance of `PgArray` to a Clojure vector, so the calls to `vec` and `.getArray` are no longer necessary.

slr71:
Thanks for the review!

PR #217, opened 2021-02-12
Update kameleon version

flynn1999:
Updated from 3.0.4 to 3.0.6.
it was 3.0.5 in [project.clj](https://github.com/cyverse-de/kameleon/compare/10d0dc6cfc7a...0e740123fa39). Does that mean it wasn't updated before?


ianmcorvidae:
Looks good. I'll merge it now.

PR #218, opened 2021-03-05
CORE-1314: added the `isBlessed` flag to app listings for DE apps pro…

slr71:
…vided that the `cyverse-blessed` attribute has been applied to the app with a value of `true`

### DIFF
--- a/src/apps/service/apps/de/listings.clj
+++ b/src/apps/service/apps/de/listings.clj
@@ -305,11 +305,19 @@
                   :value (workspace-metadata-beta-value)}]
     (set (metadata-client/filter-by-avus username app-ids [beta-avu]))))
 
+(defn- app-ids->certified-ids-set
+  "Filters the given list of app-ids into a set containing the ids of apps marked as `certified`"
+  [username app-ids]
+  (let [certified-avu {:attr  (workspace-metadata-certified-apps-attr)
+                       :value (workspace-metadata-certified-apps-value)}]
+    (set (metadata-client/filter-by-avus username app-ids [certified-avu]))))
+
 (defn- get-app-listing-formatter
   "Returns a function that can be used to format the listing for a single app. Using a higher order function for this
    makes it easier to consolidate repetitive tasks without degrating performance."
-  [user admin? perms app-ids public-app-ids]
-  (let [beta-ids-set        (app-ids->beta-ids-set (:shortUsername user) app-ids)
+  [{username :shortUsername :as user} admin? perms app-ids public-app-ids]
+  (let [beta-ids-set        (app-ids->beta-ids-set username app-ids)
+        certified-ids-set   (app-ids->certified-ids-set username app-ids)
         limit-check-results (limits/load-limit-check-results user)]
     (fn [{:keys [id] :as app}]
       (-> (assoc app :can_run (app-can-run? app))
@@ -320,6 +328,7 @@
           (format-app-permissions perms)
           (assoc :can_favor true :can_rate true :app_type "DE" :system_id c/system-id)
           (assoc :beta (contains? beta-ids-set id))
+          (assoc :isBlessed (contains? certified-ids-set id))
           (assoc :is_public (contains? public-app-ids id))
           (assoc :limitChecks (limits/format-app-limit-check-results limit-check-results app))
           (remove-nil-vals)))))
--- a/src/apps/util/config.clj
+++ b/src/apps/util/config.clj
@@ -173,6 +173,16 @@
   [props config-valid configs]
   "apps.workspace.metadata.communities.attr" "cyverse-community")
 
+(cc/defprop-optstr workspace-metadata-certified-apps-attr
+  "A metadata attr that can be applied to an app that has been reviewed and certified by DE administrators."
+  [props config-valid configs]
+  "apps.workspace.metadata.certified.attr" "cyverse-blessed")
+
+(cc/defprop-optstr workspace-metadata-certified-apps-value
+  "The value of the AVU to apply to an app that has been reviewed and certified by DE administrators."
+  [props config-valid configs]
+  "apps.workspace.metadata.certified.value" "true")
+
 (cc/defprop-str uid-domain
   "The domain name to append to the user identifier to get the fully qualified
    user identifier."
--- a/project.clj
+++ b/project.clj
@@ -30,7 +30,7 @@
                  [org.cyverse/metadata-client "3.1.1"]
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/common-cfg "2.8.1"]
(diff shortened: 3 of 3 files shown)


src/apps/util/config.clj
```diff
@@ -173,6 +173,16 @@
   [props config-valid configs]
   "apps.workspace.metadata.communities.attr" "cyverse-community")
 
+(cc/defprop-optstr workspace-metadata-certified-apps-attr
+  "A metadata attr that can be applied to an app that has been reviewed and certified by DE administrators."
+  [props config-valid configs]
+  "apps.workspace.metadata.certified.attr" "cyverse-blessed")
```
psarando, line 179: I wonder if we should prevent the non-admin service calls in https://github.com/cyverse-de/apps/blob/master/src/apps/metadata/avus.clj from setting an AVU with this attribute.

psarando (approved):
LGTM 👍

src/apps/util/config.clj
slr71: Ah, that's a good idea! I'll add it, thanks!

slr71: It looks like there's no validation for the `beta` and `communities` attributes either. I'll add checks for those as well unless anyone can think of a reason not to.

psarando: Good catch! Sounds like a good idea to me 👍

psarando (approved):
Nice! 🎉 
LGTM 👍

slr71:
Thanks for the review! I made the changes to add validation to the app metadata modification endpoints. I also added a couple of endpoints that @sri2k1us requested. (He actually requested either new endpoints or instructions for using existing endpoints. I think it makes sense to add new endpoints so that we can keep the knowledge of how apps are marked as blessed all in one place.)

These changes are pretty substantial, so if I could get a re-review, I'd appreciate it.

src/apps/metadata/avus.clj
```diff
@@ -12,15 +12,17 @@
...
 
 (defn set-avus
-  [{username :shortUsername :as user} app-id request admin?]
+  [{username :shortUsername :as user} app-id {:keys [avus] :as request} admin?]
   (let [{app-name :name :as app} (app-db/get-app app-id)]
     (validation/verify-app-permission user app "write" admin?)
-    (categorization/validate-app-name-in-hierarchy-avus username app-id app-name (:avus request))
+    (validation/validate-attrs-not-protected admin? avus)
+    (categorization/validate-app-name-in-hierarchy-avus username app-id app-name avus)
     (metadata-client/set-avus username app-id (json/encode request))))
```
slr71, line 20: It just occurred to me that there's a bug in my changes because of an initial misunderstanding I had about how this endpoint works. I don't have time to fix the bug tonight, though. I'll mark this PR as a work in progress for the time being.

slr71:
This PR should be ready for review again. The behavior of the non-admin app metadata endpoints has been changed a bit. See https://cyverse.atlassian.net/browse/CORE-1314 for details.

src/apps/metadata/avus.clj
```diff
@@ -5,22 +5,43 @@
...
+  [{username :shortUsername :as user} app-id {:keys [avus]} admin?]
   (let [{app-name :name :as app} (app-db/get-app app-id)]
     (validation/verify-app-permission user app "write" admin?)
-    (categorization/validate-app-name-in-hierarchy-avus username app-id app-name (:avus request))
-    (metadata-client/set-avus username app-id (json/encode request))))
+    (validation/validate-attrs-not-protected admin? avus)
+    (categorization/validate-app-name-in-hierarchy-avus username app-id app-name avus)
+    (let [new-avus (if admin? avus (concat avus (load-protected-avus username app-id)))]
+      (update (metadata-client/set-avus username app-id (json/encode {:avus new-avus}))
+              :avus remove-protected-avus admin?))))
```
psarando, line 36: Nice 👍 

This is good as-is, but for some reason I got tripped up for a second when I initially thought the `update` was further modifying the request getting sent to the metadata service, but I realize now it's modifying the response before passing it back to the client. Not sure if `as->` threading might help clarify the flow here 🤷 But maybe my :clojure: 👀 are just a bit rusty these days 😆

src/apps/metadata/avus.clj
```diff
@@ -5,22 +5,43 @@
...
 
 (defn update-avus
-  [{username :shortUsername :as user} app-id request admin?]
+  "Adds AVUs to an app. If `admin?` is `true` then any attribute may be added. If `admin?` is `false` then
```
psarando, line 39: Adds or updates AVUs (if the avu includes a UUID).

src/apps/routes/admin/apps.clj
```diff
@@ -70,6 +70,20 @@
...
                    (apps/admin-update-app current-user system-id (assoc body :id app-id)))))
 
+    (POST "/blessing" []
+      :query [params SecuredQueryParams]
+      :summary schema/BlessAppSummary
+      :description schema/BlessAppDescription
+      (apps/admin-bless-app current-user system-id app-id)
+      (ok))
+
+    (DELETE "/blessing" []
```
psarando, line 80: :holyhandgrenade:

src/apps/service/apps/de/listings.clj
```diff
@@ -305,11 +305,17 @@
                   :value (workspace-metadata-beta-value)}]
     (set (metadata-client/filter-by-avus username app-ids [beta-avu]))))
 
+(defn- app-ids->certified-ids-set
+  "Filters the given list of app-ids into a set containing the ids of apps marked as `certified`"
```
psarando, line 309: certified :insane:?

psarando (approved):
Very nice 🎉 We've been needing these protected AVU checks for a while, and this will let us expose app metadata to users if we wanted ✨ 

LGTM 👍

src/apps/metadata/avus.clj
slr71: Hmm, yeah, a threading macro might help here. Maybe something like this would be easier to read.

``` clojure
(-> (metadata-client/set-avus username app-id (json/encode {:avus new-avus}))
    (update :avus remove-protected-avus admin?))
```

slr71: Or the threading macro option that you mentioned:

``` clojure
(as-> (json/encode {:avus new-avus}) r
      (metadata-client/set-avus username app-id r)
      (update r :avus remove-protected-avus admin?))
```

slr71: Cool, I'll go with that one.

psarando: Yep, I was thinking of something more broken up like the second option you posted 👍

slr71:
Thanks for the reviews!

PR #222, opened 2021-04-29
CORE-1390: updated the app link in app deletion notification emails t…

slr71:
…o work with Sonora


johnworth (approved):
LGTM :shipit:

slr71:
Thanks for the review!

PR #224, opened 2021-05-13
CORE-1416: improve analysis listing performance

slr71:
This PR improves the performance of two components of analysis listing:

- listing permissions for all analyses that a user may access.
- listing representative job steps for the analyses in the listing

For the time being, the permissions listing improvement is achieved by eliminating some of the unnecessary data from the permission listing. In one case where the timeout was occurring, the user had access to more than 15000 analyses. Serializing and deserializing that many permissions was taking a significant amount of time, so a new endpoint was created for an abbreviated permission listing that omits some of the data that wasn't being used in this case. This change shortened the response time by a little more than a second in most cases.

After the database merge is complete, we're probably going to want to refactor this code yet again in order to combine the analysis listing with the permission lookup. That should save a significant amount of time as well.

The improvement to the representative job steps listing was achieved by refactoring the database query. The new database query has the potential to return more rows than we actually need, but this won't do any harm because this query is never used to determine which analyses (or steps) a user sees.

There is still some room for improvement. For example, the code to obtain the status information for each batch takes about a tenth of a second for each batch in the listing, which can slow things down quite a bit. Combining this into the query itself is likely to help a little bit. I think that we've made a big enough improvement to prevent most response timeouts for analysis listings for the time being, though. We should be able to safely wait until we have more time to implement additional improvements.

### DIFF
--- a/src/apps/clients/data_info.clj
+++ b/src/apps/clients/data_info.clj
@@ -63,7 +63,8 @@
   (->> (slurp (get-file-contents user path))
        (string/split-lines)
        (remove empty?)
-       (drop 1)))
+       (drop 1)
+       (mapv string/trim)))
 
 (defn create-directory
   [user path]
--- a/src/apps/clients/permissions.clj
+++ b/src/apps/clients/permissions.clj
@@ -181,9 +181,10 @@
 (def unshare-tool (partial unshare-resource (rt-tool)))
 
 (defn- get-public-resource-ids [resource-type]
-  (->> (pc/get-subject-permissions-for-resource-type (client) "group" (ipg/grouper-user-group-id) resource-type false)
+  (->> (pc/get-abbreviated-subject-permissions-for-resource-type
+        (client) "group" (ipg/grouper-user-group-id) resource-type false)
        :permissions
-       (map (comp uuidify :name :resource))
+       (map (comp uuidify :resource_name))
        set))
 
 (def get-public-app-ids (partial get-public-resource-ids "app"))
--- a/src/apps/clients/permissions.clj
+++ b/src/apps/clients/permissions.clj
@@ -31,6 +31,11 @@
   [perms]
   (into {} (map (juxt (comp uuidify :name :resource) :permission_level) perms)))
 
+(defn- group-abbreviated-permissions
+  "Groups abbreviated permissions by resource ID. The resource ID must be a UUID."
+  [perms]
+  (into {} (map (juxt (comp uuidify :resource_name) :permission_level) perms)))
+
 (defn extract-error-message
   [body]
   ((some-fn :reason :message) (service/parse-json body)))
@@ -43,18 +48,18 @@
 (def register-private-analysis (partial register-private-resource (rt-analysis)))
 (def register-private-tool (partial register-private-resource (rt-tool)))
 
-(defn- filter-perms-response
+(defn- filter-abbreviated-perms-response
   [response filter-fn]
-  (group-permissions (filter filter-fn (:permissions response))))
+  (group-abbreviated-permissions (filter filter-fn (:permissions response))))
 
 (defn- load-resource-permissions*
   ([resource-type user filter-fn]
-   (filter-perms-response
-    (pc/get-subject-permissions-for-resource-type (client) "user" user resource-type true)
+   (filter-abbreviated-perms-response
+    (pc/get-abbreviated-subject-permissions-for-resource-type (client) "user" user resource-type true)
     filter-fn))
   ([resource-type user min-level filter-fn]
-   (filter-perms-response
-    (pc/get-subject-permissions-for-resource-type (client) "user" user resource-type true min-level)
+   (filter-abbreviated-perms-response
+    (pc/get-abbreviated-subject-permissions-for-resource-type (client) "user" user resource-type true min-level)
     filter-fn)))
 
 (defn- resource-id-filter
--- a/src/apps/persistence/jobs.clj
+++ b/src/apps/persistence/jobs.clj
@@ -673,7 +673,7 @@
                             (h/where [:= :p.id :c.parent_id]))])))
 
 (defn- representative-job-ids-query
-  "Returns a query that obtains the ID, parent ID, and representative job ID of every job in the related_job_ids query,
+  "Returns a query that obtains the ID, parent ID, and representative job ID of every job in the related-job-ids-alias query,
   which should be available as a CTE. The representative job ID is the job ID whose steps represent the steps of every
   job in a batch. For individual jobs, the representative job ID is the job ID itself. For batch jobs, the
   representative job ID is the ID of any of the child jobs in the batch."
--- a/src/apps/persistence/jobs.clj
+++ b/src/apps/persistence/jobs.clj
@@ -688,8 +688,8 @@
                        (h/where [:= :c.parent_id :p.id])
                        (h/limit 1))
                    :representative_id])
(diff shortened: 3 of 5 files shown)


slr71:
I think I might have found a bug in this. I'm going to mark this PR as a draft until I can either fix the problem or convince myself that it doesn't really exist.

slr71:
Okay, I updated the query again. The query that was being used yesterday had a bit of a problem because it was listing the steps for every job within every batch in the result. I'd thought that this wasn't a problem before yesterday, but I hadn't really thought it through; the number of jobs in a single batch can be quite large, so while the analysis listing itself would have been correct, we would have been doing a lot of unnecessary processing.

There was another problem that existed before my earlier changes. We were grouping the representative job steps first by parent ID then by batch ID, which meant that jobs in a batch appeared to have no representative steps. Once again, I don't think that caused any problem in the listing because the representative job steps are only being used to determine whether or not each job can be shared. Since jobs in a batch cannot be shared individually (the entire batch would have to be shared instead), jobs in a batch would have been marked as not sharable anyway. This still would have caused confusion if we had to use the list of representative job steps for a different purpose, though.

The new query removes the need to group by parent ID, does not return rows related to analyses that are not in the listing, and runs faster than both the original query and the one that was in place yesterday.

src/apps/persistence/jobs.clj
```diff
@@ -645,23 +650,73 @@
...
+                :s.external_id
+                :s.start_date
+                :s.end_date
+                :s.status
+                [:t.name :job_type]
+                :s.app_step_number)
+      (h/from [:representative_job_ids :r])
+      (h/join [:job_steps :s] [:= :r.representative_id :s.job_id]
+              [:job_types :t] [:= :s.job_type_id :t.id])
+      hsql/format))
```
slr71, line 711: I _really_ like HoneySQL.

src/apps/service/apps/job_listings.clj
```diff
@@ -41,8 +41,8 @@
 (defn- job-supports-sharing?
   [apps-client perms rep-steps {:keys [parent_id id]}]
   (and (nil? parent_id)
-       (job-permissions/job-steps-support-job-sharing? apps-client (rep-steps id))
-       (= (get perms id) "own")))
+       (= (get perms id) "own")
+       (job-permissions/job-steps-support-job-sharing? apps-client (rep-steps id))))
```
slr71, line 45: I switched the order of the conditions here so that the simpler comparison would happen first.

src/apps/clients/permissions.clj
```diff
@@ -31,6 +31,11 @@
   [perms]
   (into {} (map (juxt (comp uuidify :name :resource) :permission_level) perms)))
 
+(defn- group-abbreviated-permissions
+  "Groups abbreviated permissions by resource ID. The resource ID must be a UUID."
+  [perms]
+  (into {} (map (juxt (comp uuidify :resource_name) :permission_level) perms)))
+
```
slr71, line 38: The changes in this file are all related to the new endpoint in the `permissions` service.

src/apps/service/apps/job_listings.clj
```diff
@@ -70,7 +70,8 @@
   [{user-id :user_id :as job} rep-steps]
   (let [interactive? (fn [step] (= (:job_type step) jp/interactive-job-type))
         get-url      (fn [step] (str (interapps-url (url (config/interapps-base)) user-id (:external_id step))))]
-    (seq (map get-url (filter interactive? (rep-steps (:id job)))))))
+    (when-not (:is_batch job)
+      (seq (map get-url (filter interactive? (rep-steps (:id job))))))))
```
slr71, line 74: With the previous query, no interactive steps could ever have been associated with a batch parent job (even if interactive batches were allowed). With the new query, interactive steps would appear in a batch containing interactive jobs, so checking to see if the job is a batch first prevents batch parent jobs from having interactive URLs associated with them. Admittedly, interactive jobs still aren't allowed in batches, but I figured that it wouldn't hurt to future-proof this code a bit.

src/apps/service/apps/job_listings.clj
```diff
@@ -108,7 +109,7 @@
         search-params    (util/default-search-params params :startdate default-sort-dir)
         types            (.getJobTypes apps-client)
         jobs             (list-jobs* user search-params types analysis-ids)
-        rep-steps        (group-by (some-fn :parent_id :job_id) (jp/list-representative-job-steps (mapv :id jobs)))
+        rep-steps        (group-by :job_id (jp/list-representative-job-steps (mapv :id jobs)))
```
slr71, line 112: The new query eliminates the need to group by the parent ID.

src/apps/clients/data_info.clj
```diff
@@ -63,7 +63,8 @@
   (->> (slurp (get-file-contents user path))
        (string/split-lines)
        (remove empty?)
-       (drop 1)))
+       (drop 1)
+       (mapv string/trim)))
```
slr71, line 67: This change isn't directly related to `CORE-1416`. This fixes a problem where trailing tabs in path list files were causing jobs to fail.

src/apps/persistence/jobs.clj
```diff
@@ -645,23 +650,73 @@
...
+        (h/from related-job-ids-alias)
+        (h/where [:not-in :id (-> (h/select :id) (h/from batch-parent-job-ids-alias))]))
+    (-> (h/select :id :parent_id
+                  [(-> (h/select :c.id)
+                       (h/from [related-job-ids-alias :c])
+                       (h/where [:= :c.parent_id :p.id])
+                       (h/limit 1))
+                   :representative_id])
+        (h/from [:related_job_ids :p])
+        (h/where [:in :id (-> (h/select :id) (h/from :batch_parent_job_ids))]))]})
```
psarando, line 692: I'm not as familiar with HoneySQL yet, but should these be the `alias` params passed into this function, rather than the raw keys?
```suggestion
        (h/from [related-job-ids-alias :p])
        (h/where [:in :id (-> (h/select :id) (h/from batch-parent-job-ids-alias))]))]})
```

src/apps/persistence/jobs.clj
```diff
@@ -645,23 +650,73 @@
...
+  as a common table expression (CTE). This exists just to make the representative_job_ids query a little cleaner."
+  [related-job-ids-alias]
+  (-> (h/select :id)
+      (h/from [related-job-ids-alias :p])
+      (h/where [:exists (-> (h/select :*)
+                            (h/from [related-job-ids-alias :c])
+                            (h/where [:= :p.id :c.parent_id]))])))
+
+(defn- representative-job-ids-query
+  "Returns a query that obtains the ID, parent ID, and representative job ID of every job in the related_job_ids query,
```
psarando, line 676: Should this doc refer to the `related-job-ids-alias` param instead?
```suggestion
  "Returns a query that obtains the ID, parent ID, and representative job ID of every job in the related-job-ids-alias query,
```

psarando (approved):
Looks good! 👍

I don't think I caught any issues, though I had a question about the `representative-job-ids-query` function.

src/apps/persistence/jobs.clj
slr71: Ah, yes. I meant to change that, but I must have missed it. Thanks for catching that.

src/apps/persistence/jobs.clj
slr71: Yep, it should! Thanks for catching that.

slr71:
Thanks for the review!

PR #227, opened 2021-06-11
CORE-1445: switch from a namespace alias to the full name in a namesp…

slr71:
…aced keyword


slr71:
This change is tiny; I'm going to merge it now. Reviews are still welcome!

PR #230, opened 2021-09-08
CORE-1489: be smarter about fetching docker registry authentication information

ianmcorvidae:
Instead of taking everything up to the last slash, split on slashes, then repeatedly check for registry information starting with the full name, then removing slash-separated parts starting from the end. This way, it should work for any slash-separated prefix match.

Would appreciate some extra eyes on this since I'm feeling like I missed something in the logic, but anyway.

### DIFF
--- a/src/apps/persistence/docker_registries.clj
+++ b/src/apps/persistence/docker_registries.clj
@@ -2,7 +2,8 @@
   (:use [apps.persistence.entities :only [docker-registries]]
         [apps.util.db :only [transaction]]
         [korma.core :exclude [update]])
-  (:require [korma.core :as sql]))
+  (:require [korma.core :as sql]
+            [clojure.string :as string]))
 
 (defn get-registries
   []
@@ -12,6 +13,14 @@
   [name]
   (first (select docker-registries
                  (where {:name name}))))
+
+(defn get-registry-from-image
+  [image-name]
+  (let [parts (string/split image-name #"/")]
+    (loop [n (count parts)]
+      (if-let [reg (get-registry (string/join "/" (take n parts)))]
+        reg
+        (recur (- n 1))))))
 
 (defn add-registry
   [name username password]
--- a/src/apps/persistence/docker_registries.clj
+++ b/src/apps/persistence/docker_registries.clj
@@ -18,9 +18,10 @@
   [image-name]
   (let [parts (string/split image-name #"/")]
     (loop [n (count parts)]
-      (if-let [reg (get-registry (string/join "/" (take n parts)))]
-        reg
-        (recur (- n 1))))))
+      (when-not (= n 0)
+        (if-let [reg (get-registry (string/join "/" (take n parts)))]
+          reg
+          (recur (- n 1)))))))
 
 (defn add-registry
   [name username password]
--- a/src/apps/containers.clj
+++ b/src/apps/containers.clj
@@ -9,7 +9,7 @@
                                           ports
                                           data-containers
                                           interapps-proxy-settings]]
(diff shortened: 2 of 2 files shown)


src/apps/persistence/docker_registries.clj
```diff
@@ -13,6 +14,15 @@
   (first (select docker-registries
                  (where {:name name}))))
 
+(defn get-registry-from-image
+  [image-name]
+  (let [parts (string/split image-name #"/")]
+    (loop [n (count parts)]
+      (when-not (= n 0)
+        (if-let [reg (get-registry (string/join "/" (take n parts)))]
```
ianmcorvidae, line 22: `get-registry` being inside this loop does mean this can do up to `(count parts)` queries against the DB. I'm wondering if that's worth fixing or not; I guess it would most affect deeply-nested image names that don't require auth (e.g. something like `foo.bar.baz/project/subproject/subsubproject/image` would query five times)

src/apps/persistence/docker_registries.clj
```diff
@@ -13,6 +14,15 @@
...
+          reg
+          (recur (- n 1)))))))
```
psarando, line 24: Maybe it's safe to assume that the registry will usually not have a `/` in the name?
Then we could loop the other way, progressively adding parts until we find the registry name, preventing us hitting the db too many times in most cases?
```suggestion
    (loop [n 1]
      (when (> (count parts) n)
        (if-let [reg (get-registry (string/join "/" (take n parts)))]
          reg
          (recur (+ n 1)))))))
```

ianmcorvidae: I didn't do this because I guess I was figuring we wanted the longest match, in order to allow greater specificity with a default (e.g. one set of credentials for dockerhub generally, but a different one for a specific username's images, or such). That might be another question though, if I'm overengineering it :)

I suppose if we're matching at the start, I might be able to roll it all directly into one SQL query, something in the vein of `SELECT ... WHERE $image_name LIKE name || '%'` maybe

psarando: I see, I didn't know that was a possibility.
In that case, this looks good to me.

psarando (approved):
LGTM 👍

ianmcorvidae:
Cool, thanks for the review. I'll merge it now and then testing can hopefully happen. Tomorrow, possibly.

slr71 (commented):
~It looks like there's some other code that calls `get-registry`:~

[Edit: I went to change what I'd written and accidentally sent the comment.]

The trusted registry code also makes the assumption that the registry name won't contain any slashes. It may be necessary to add support for registry names containing slashes there as well:

```
src/apps/service/apps/de/listings.clj:        get-registry           (comp first #(string/split % #"/" 2) :image_name)
src/apps/service/apps/de/listings.clj:        in-untrusted-registry? #(not (contains? trusted-registries (get-registry %)))]
src/apps/service/apps/de/validation.clj:        get-registry           (comp first #(string/split % #"/" 2) :image_name)
src/apps/service/apps/de/validation.clj:        in-untrusted-registry? #(not (contains? trusted-registries (get-registry %)))]
```

ianmcorvidae:
Yeah, that would probably be good, though at least that code won't have the issue this code did (despite the same name, it's not the same `get-registry`, and the apps.service.* code all grabs the text before the first slash, not up-to-the-last-slash like the prior code for auth stuff). I would like to add support for full-fledged prefix matching to the trusted registry code so we can trust certain projects on harbor but not others, though.

slr71:
Yeah, that definitely doesn't have to be done with this PR.

PR #231, opened 2021-09-10
CORE-1482 Add cyverse-blessed app category

ash3rz:


src/apps/service/apps/de/listings.clj
```diff
@@ -134,13 +140,32 @@
...
+  [_ workspace params]
+   {:system_id de-system-id
+    :id        cyverse-blessed-id
+    :name      "CyVerse Blessed"
+    :is_public false
+    :total     (count-apps-for-user nil (:id workspace) {:app-ids (:listing-ids params)})})
+
+(defn list-cyverse-blessed-apps
+  [user workspace params]
+   (let [blessed-ids (app-ids->certified-ids-set (:username user) (:app-ids params))]
```
ash3rz, line 154: This seems to work because `list-apps-in-group` already makes a call out to permissions `load-app-permissions` which returns the list of all apps the user has permission to view.  Then it calls `augment-listing-params` to add that list as `:app-ids` to `params`.  So here, I'm just taking that list of `:app-ds` and filtering it down to the blessed ids.

src/apps/service/apps/de/listings.clj
```diff
@@ -400,8 +420,9 @@
   [user workspace group-id perms {:keys [public-app-ids] :as params}]
   (when-let [format-fns (virtual-group-fns group-id)]
     (let [app-listing ((:format-listing format-fns) user workspace params)
-          format-app  (get-app-listing-formatter user false perms (map :id app-listing) public-app-ids)]
-      (-> ((:format-group format-fns) user workspace params)
+          app-ids     (map :id app-listing)
+          format-app  (get-app-listing-formatter user false perms app-ids public-app-ids)]
+      (-> ((:format-group format-fns) user workspace (assoc params :listing-ids app-ids))
```
ash3rz, line 425: Adding the listing ids to `params` was the only way I could think of to get the total count for formatting. The other virtual categories seem to have their own DB query that they run to get the count, but I didn't think I could do that since I need the blessed-ids from permissions and I already have that list here anyway.  Very open to suggestions if this seems off.

sri2k1us:
Sorry. Little late. This will be called `Featured Apps` in the UI. Should we just use that terminology here as well ?

ash3rz:
> Sorry. Little late. This will be called `Featured Apps` in the UI. Should we just use that terminology here as well ?

Sure, I can update the name here.

src/apps/service/apps/de/listings.clj
```diff
@@ -134,13 +140,32 @@
...
+  [user _ params]
+   {:system_id de-system-id
+    :id        featured-apps-id
+    :name      "Featured Apps"
+    :is_public false
+    :total     (future (count-apps-for-user nil nil {:app-ids (app-ids->certified-ids-set (:username user) (:app-ids params))}))})
+
+(defn list-featured-apps
+  [user workspace params]
+   (let [blessed-ids (app-ids->certified-ids-set (:username user) (:app-ids params))]
```
psarando, line 154: Ideally we wouldn't have to hit the metadata service twice per listing to get the filtered list of blessed IDs.
Unfortunately I don't see another precedent for doing that with these virtual groups.

Maybe we could add an `augment-virtual-group-listing-params` function which could augment the listing params once more to update the `:app-ids` param with the filtered IDs, but for now it would only do that if the group ID is this Featured Apps category. It could be called in `list-apps-in-virtual-group`, just before the listing and format functions are called. Then both the listing and format functions could just use `(:app-ids params)` 🤔 

Actually, we should probably pass the entire `params` map to the count function? It could make the total count more accurate, since it could be affected by other listing params that might filter the listing even further. For example, if the user wants to view only VICE apps in the Featured category.

psarando (commented):
This looks right so far, though I wonder if we can do the listing by only hitting the metadata service once per listing.

I'm also not sure if the listing total will be correct with other potential filtering params.

src/apps/service/apps/de/listings.clj
ash3rz: Ah, I think you're correct about the count.  Thanks for catching that!

I actually started to do what you're suggesting but added a `:listing-ids` param to params so anything referencing `:app-ids` wouldn't be affected, but ran into issues so I reverted that.  If I try augmenting the params before any of the listing or format functions are called though like you suggested, it might work. I'll give that a shot!

psarando (approved):
LGTM 👍 ✨

slr71:
Looks good to me! :shipit:

src/apps/routes/apps/categories.clj
```diff
@@ -32,6 +33,14 @@
     :description schema/AppCategoryListingDocs
     (ok (apps/get-app-categories current-user params)))
 
+  (GET "/featured" []
```
psarando, line 36: 👍

ash3rz:
Thanks for the reviews!

PR #232, opened 2021-10-14
CORE-1500 Add `pendingRequest` key to limit checks

ash3rz:
This will add a boolean to the current limit checks to allow the UI to know if the user already has a VICE access request that is awaiting admin review.


psarando (approved):
LGTM 👍

ash3rz:
Thanks for the review!

PR #236, opened 2021-11-17
CORE-1503 Update app listings and search queries for versions

psarando:
This PR will update app listings and search queries for versions.

This PR should not be merged until after cyverse-de/de-database#10 and cyverse-de/common-swagger-api#75 are merged.


ash3rz (approved):
Not my area of expertise, but this aligns with what I remember from the design discussion 👍

PR #238, opened 2021-12-01
CORE-1634: allow analysis listings and analysis stat listings to incl…

slr71:
…ude deleted analyses

### DIFF
--- a/src/apps/service/apps/job_listings.clj
+++ b/src/apps/service/apps/job_listings.clj
@@ -48,7 +48,7 @@
   [{:keys [parent_id id] :as job}]
   (remove-nil-vals
    {:app_description (:app_description job)
-    :app_id          (:app_id job)
+    :app_id          (or (:app_id job) "unknown")
     :app_name        (:app_name job)
     :description     (:description job)
     :enddate         (job-timestamp (:end_date job))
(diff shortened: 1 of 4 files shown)


psarando (approved):
LGTM 👍

src/apps/service/apps/job_listings.clj
```diff
@@ -48,7 +48,7 @@
   [{:keys [parent_id id] :as job}]
   (remove-nil-vals
    {:app_description (:app_description job)
-    :app_id          (:app_id job)
+    :app_id          (or (:app_id job) "unknown")
```
ianmcorvidae, line 51: You may have already, but if you haven't, we should check that anywhere that allows an app ID parameter that flows from this function (or any others changed to allow non-UUID values) won't choke on `"unknown"`. Might be fine because of Agave jobs, but sometimes we have things special-cased, or endpoints that only apply to DE apps, so I think it's worth a review.

ianmcorvidae (approved):
One thing to check but the change looks reasonable regardless (if something needs changed due to the comment it's probably in common-swagger-api anyway, heh)

src/apps/service/apps/job_listings.clj
slr71: Yep, it's safe precisely because of Agave jobs. I double-checked yesterday before I made the change.

slr71:
Thanks for the reviews!

PR #239, opened 2021-12-03
CORE-1506 Update app details endpoint to use latest app version

psarando:
This PR will update the app details endpoint to use info from the latest app version, which includes updates to allow the `GET /apps/{system-id}/{app-id}/documentation` endpoint to use docs from the latest app version.

This PR is on top of #237, so only the last 2 commits need to be reviewed for this one.

Summary of changes:

* Added `app_versions` table references to `apps.persistence.entities` and removed an unused `rating_listing` entity.
* Added `apps.persistence.app-metadata/get-app-latest-version`
* Refactored a few functions to use `apps.service.apps.de.docs/has-docs?` so that it can check for docs in the latest app version.
* Removed unused `apps.service.apps.de.docs/upsert-app-docs` function.
* Updated functions for getting app docs and references to use an app version ID.
* Added latest app version ID to app documentation responses.

PR #240, opened 2021-12-04
CORE-1625 CORE-1626 Update add/edit app docs to use latest app version

psarando:
This PR will update the functions for adding and editing app docs to add/edit docs in the latest app version.

The work for these functions were closely related, and affected the same files, so I combined them in a single commit for a single PR.

This PR is on top of #239, so only the last commit needs review for this PR.


psarando:

PR #241, opened 2021-12-08
CORE-1508 Update app UI endpoint to use latest app version

psarando:
This PR will update the `GET /apps/{system-id}/{app-id}/ui` endpoint to use latest app version.

This PR is on top of #240, so only the last commit needs review.


psarando:

PR #242, opened 2021-12-09
CORE-1637 Update app relabel endpoint to use latest app version

psarando:
This PR will update the `PATCH /apps/:system-id/:app-id` endpoint to use the latest version of an app.

This PR is on top of #241, so only the last commit needs review.

### DIFF
--- a/src/apps/persistence/app_metadata.clj
+++ b/src/apps/persistence/app_metadata.clj
@@ -105,6 +105,13 @@
   [app-id]
   (assert-not-nil [:app-id app-id] (app-listing/get-app-listing (uuidify app-id))))
 
+(defn get-app-latest-version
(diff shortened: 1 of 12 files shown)


src/apps/persistence/app_metadata.clj
```diff
@@ -532,11 +531,6 @@
                                             {:output parameter-id})))
    (remove-workflow-map-orphans)))
 
-(defn update-app-labels
-  "Updates the labels in an app."
-  [req]
-  (relabel/update-app-labels req))
```
psarando, line 528: This indirection was removed since `apps.persistence.app-metadata.relabel/update-app-labels` now needs to reference `apps.persistence.app-metadata/get-app-latest-version`, which was causing a circular dependency.

psarando:

PR #243, opened 2021-12-17
CORE-1638 Update PUT /apps/:system-id/:app-id to use latest version

psarando:
This PR will update the `PUT /apps/:system-id/:app-id` endpoint to use and update the latest app version.

This PR is on top of #242, so only the last commit needs review.

PR #244, opened 2021-12-17
CORE-1656 Update PATCH /admin/apps/:system-id/:app-id for app versions

psarando:
This PR will update the `PATCH /admin/apps/:system-id/:app-id` endpoint to update the latest app version, and to mark all versions of an app as deleted or disabled.

This also updates the delete behavior for the `DELETE [/admin]/apps/:system-id/:app-id` endpoints, which will also now mark all versions of an app as deleted.

This PR is on top of #243, so only the last 2 commits need review.


psarando:

PR #245, opened 2022-01-07
CORE-1639 Update add-app to add new version info

psarando:
This PR will update the `POST /apps/:system-id` endpoint to add new version info for new apps.

This PR is on top of #244, so only the last commit needs review.


psarando:

PR #247, opened 2022-01-20
Update common-swagger-api to 3.2.3

johnworth:
Updates the version of common-swagger-api. This should enable analysis submissions to include the max_cpu_cores field. That doesn't mean the service does anything with the field yet, that's still being looked at.


psarando (approved):
LGTM :shipit:

PR #250, opened 2022-05-19
CORE-1741 Use client max CPU if lower than tool default

psarando:
The `max_cpu_cores` request submitted from the client was ignored and not used in submissions to the job services.

This PR will update the job submission endpoint use the `max_cpu_cores` request submitted from the client, as long as it's less than the tool's default restriction.

Will also use max `memory_limit` from the client in the same way, if the endpoint is updated in the future to allow it in requests.

### DIFF
--- a/src/apps/service/apps/de/jobs/common.clj
+++ b/src/apps/service/apps/de/jobs/common.clj
@@ -58,6 +58,11 @@
                   :min_cpu_cores
                   :min_disk_space]))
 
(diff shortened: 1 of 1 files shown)


johnworth (approved):
Seems like a logical change. :shipit:

slr71 (approved):
LGTM :shipit:

psarando:
Thanks for the reviews 👍

PR #246, opened 2022-01-11
CORE-316 Initial updates for app versions db migrations

psarando:
This PR will update existing endpoints for compatibility with app versions db migrations.

This PR should not be merged until after cyverse-de/de-database#10 and cyverse-de/common-swagger-api#75 are merged.

This PR will update the following apps endpoints to use, update, create, and/or return info for the latest app version.

* `GET /apps`
* `POST /apps/{system-id}`
* `DELETE /apps/{system-id}/{app-id}`
* `GET /apps/{system-id}/{app-id}`
* `PATCH /apps/{system-id}/{app-id}`
* `PUT /apps/{system-id}/{app-id}`
* `POST /apps/{system-id}/{app-id}/copy`
* `GET /apps/{system-id}/{app-id}/listing`
* `GET /apps/{system-id}/{app-id}/details`
* `GET /apps/{system-id}/{app-id}/documentation`
* `PATCH /apps/{system-id}/{app-id}/documentation`
* `POST /apps/{system-id}/{app-id}/documentation`
* `GET /apps/{system-id}/{app-id}/integration-data`
* `GET /apps/{system-id}/{app-id}/is-publishable`
* `GET /apps/{system-id}/{app-id}/tasks`
* `GET /apps/{system-id}/{app-id}/tools`
* `GET /apps/{system-id}/{app-id}/ui`
* `POST /apps/pipelines`
* `PUT /apps/pipelines/{app-id}`
* `POST /apps/pipelines/{app-id}/copy`
* `GET /apps/pipelines/{app-id}/ui`
* `GET /admin/apps`
* `DELETE /admin/apps/{system-id}/{app-id}`
* `PATCH /admin/apps/{system-id}/{app-id}`
* `GET /admin/apps/{system-id}/{app-id}/details`
* `PATCH /admin/apps/{system-id}/{app-id}/documentation`
* `POST /admin/apps/{system-id}/{app-id}/documentation`

PR #253, opened 2022-06-29
CORE-1749 Fix admin integration-data update endpoint for app versions

psarando:
This PR will update the `PUT /admin/apps/:system-id/:app-id/integration-data/:integration-data-id` endpoint so that it will update an app's latest version.

This should have been part of the changes for #246.


psarando:
I'll go ahead and merge this now, but reviews still welcome.

PR #254, opened 2022-06-29
CORE-1750 Fix admin integration-data delete endpoint for app versions

psarando:
This PR will fix the `DELETE /admin/integration-data/:integration-data-id` endpoint so that it will check the `app_versions` table for integration data in use by apps.

This should have been part of the changes for #246.


psarando:
I'll go ahead and merge this now, but reviews still welcome.

PR #255, opened 2022-09-30
CORE-1534 Add admin app version endpoints

psarando:
This PR will add admin app version details and update endpoints.

It will also update the sharing endpoint to only share tools in undeleted versions, and the publish endpoints to only validate and publish tools in undeleted versions.

This change now requires admin undelete actions for app versions to ensure tools are not private or missing, and also ensures workflows are using tasks from public apps; and these validation checks are also now in the create version endpoints for public apps and pipelines.


ianmcorvidae (approved):
I didn't catch anything!

psarando:
Thanks for the review! 👍

PR #256, opened 2022-10-14
CORE-1533 Add admin sharing endpoints

psarando:
This PR will add `/admin/apps/sharing` and `/admin/apps/unsharing` endpoints, which will allow admins to grant `write` permissions to integrators who wish to create new versions of their public apps.

The essence of this PR was to bypass the `own` permission checks in `apps.service.apps.de.sharing` based on an `admin?` flag, and adding admin versions of the sharing endpoints that injects a `true` value for this flag down through the sharing functions.


johnworth (approved):
I don't see anything wrong here, :shipit:

psarando:
Thanks for the review 👍 

I'll merge this now, but additional reviews still welcome 🙇

PR #258, opened 2022-10-18
CORE-316 Update app version create endpoint to copy documentation

psarando:
This PR will update the app version creation endpoint (CORE-1510) and the pipeline creation endpoint (CORE-1756) to copy documentation from the previous version's documentation.


psarando:
I think this is pretty straight forward, so I'll go ahead and merge it now, but reviews are still welcome.

PR #259, opened 2022-11-16
CORE-1808 Return overall_job_type field in app listings and details

psarando:
Updated the `apps.service.apps.de.listings/format-app-pipeline-eligibility` function to no longer remove the `overall_job_type` field in app listing details.

This will allow the UI to label apps with the `interactive` job type as VICE apps.


psarando:
Thanks for the review! 👍

PR #260, opened 2023-01-24
CORE-1871: fix an error that was occurring when DE administrators app…

slr71:
…rove app publication requests

The DE admin UI sends an empty request body by default when an administrator approves an app publication request. The simplest option at this time is to update the `apps` service so that it doesn't attempt to update the app if none of the relevant fields are included in the request body.


psarando (approved):
LGTM 👍 :shipit:

johnworth (approved):
LGTM

slr71:
Thanks for the speedy reviews! 💨

PR #261, opened 2023-02-14
CORE-1875: improved the performance of the analysis listing queries.

slr71:
The primary purpose of this change is to improve the performance of the analysis listing queries. It was easier to do this using HoneySQL than Korma, so this change also rewrites the SQL generation code so that it uses HoneySQL.


src/apps/persistence/job_limits.clj
```diff
@@ -6,13 +6,13 @@
...
-            [honeysql.core :as sql]
-            [honeysql.helpers :as h]))
+            [honey.sql :as sql]
+            [honey.sql.helpers :as h]))
 
 (defn- count-concurrent-vice-jobs-query
   "Builds a query to list the concurrently running VICE jobs for a user."
   [username]
-  (-> (h/select (sql/call :count (sql/raw "DISTINCT j.id")))
+  (-> (h/select [[:count [:distinct :j.id]]])
```
slr71, line 15: These were some nice improvements in HoneySQL version 2.

src/apps/persistence/jobs.clj
```diff
@@ -106,52 +109,39 @@
...
-(defn- filter-field->where-field
-  "Returns a field key for use in a job query where-clause map, based on the given filter field."
-  [field]
-  (case field
-    "app_name" (sqlfn :lower :j.app_name)
-    "name"     (sqlfn :lower :j.job_name)
-    (keyword (str "j." field))))
-
-(defn- filter-map->where-clause
+(defn- hsql-filter-map->where-clause
```
slr71, line 111: I was on the fence about removing the `hsql-` prefixes from these functions. I thought I might leave them in for now just to clearly indicate that these functions use HoneySQL rather than Korma.

src/apps/persistence/jobs.clj
```diff
@@ -680,9 +723,9 @@
                             (h/where [:= :p.id :c.parent_id]))])))
 
 (defn- representative-job-ids-query
-  "Returns a query that obtains the ID, parent ID, and representative job ID of every job in the related-job-ids-alias query,
-  which should be available as a CTE. The representative job ID is the job ID whose steps represent the steps of every
-  job in a batch. For individual jobs, the representative job ID is the job ID itself. For batch jobs, the
+  "Returns a query that obtains the ID, parent ID, and representative job ID of every job in the related-job-ids-alias
+  query, which should be available as a CTE. The representative job ID is the job ID whose steps represent the steps of
+  every job in a batch. For individual jobs, the representative job ID is the job ID itself. For batch jobs, the
```
slr71, line 727: My editor was complaining about the line lengths here.

ianmcorvidae (approved):
Eyes are crossing a little bit at the very similar code in different DSLs, but I think this looks reasonable to me. I guess the main functional change is adding in the various accessible resource CTEs, which looks pretty good. Feels like all this SQL could still be a lot cleaner but it always does and moving to HoneySQL will hopefully facilitate more improvements.

psarando (approved):
LGTM 👍 ✨

src/apps/service/apps/job_listings.clj
```diff
@@ -87,32 +88,29 @@
...
           :interactive_urls (interactive-urls job rep-steps))))
 
-(defn- list-jobs*
-  [{:keys [username]} search-params types analysis-ids]
-  (jp/list-jobs-of-types username search-params types analysis-ids))
-
 (defn- count-jobs
   [{:keys [username]} {:keys [filter include-hidden include-deleted]} types analysis-ids]
-  (jp/count-jobs-of-types username filter include-hidden include-deleted types analysis-ids))
+  (jp/hsql-count-jobs-of-types username filter include-hidden include-deleted types analysis-ids))
```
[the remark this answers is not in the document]
slr71: Ah, good catch! I'll update that. Thanks!

slr71:
Thanks for the reviews!

PR #263, opened 2023-02-24
Enable sharing analysis inputs and outputs with groups

ianmcorvidae:
The `get-user-from-subject` function is copied from terrain, and other than that it's just using it.

I still need to test this, but reviews are welcome as I do.

### DIFF
--- a/src/apps/clients/async_tasks.clj
+++ b/src/apps/clients/async_tasks.clj
@@ -2,6 +2,7 @@
   (:require [apps.util.config :as config]
             [clojure.string :as string]
             [clojure.tools.logging :as log]
+            [otel.otel :as otel]
             [async-tasks-client.core :as async-tasks-client]))
 
 (defn get-by-id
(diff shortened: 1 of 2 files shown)


src/apps/clients/async_tasks.clj
```diff
@@ -34,9 +35,13 @@
...
-  (let [^Runnable task-thread (fn [] (thread-function async-task-id))]
-    (.start (Thread. task-thread (str prefix "-" (string/replace async-task-id #".*/tasks/" "")))))
-  async-task-id)
+  (otel/with-span [outer-span ["run-async-thread" {:kind :producer :attributes {"async-task-id" (str async-task-id)}}]]
+    (let [^Runnable task-thread (fn []
+                                  (with-open [_ (otel/span-scope outer-span)]
+                                    (otel/with-span [s ["async thread" {:kind :consumer :attributes {"async-task-id" (str async-task-id)}}]]
+                                      (thread-function async-task-id))))]
+      (.start (Thread. task-thread (str prefix "-" (string/replace async-task-id #".*/tasks/" "")))))
+    async-task-id))
```
ianmcorvidae, line 44: This is basically just copied from data-info. It _should_ mean the async threads get picked up better by opentelemetry.

src/apps/service/apps/jobs/sharing.clj
```diff
@@ -148,7 +155,8 @@
 
 (defn- process-job-inputs
   [f apps-client job]
-  (remove nil? (map f (list-job-inputs apps-client job))))
+  (otel/with-span [s ["process-job-inputs"]]
+    (doall (remove nil? (map f (list-job-inputs apps-client job))))))
```
ianmcorvidae, line 159: When I was testing, my input file wasn't getting shared. Adding this `doall` fixed it -- I think that the function wasn't actually getting called for inputs properly. Also added it to `process-child-jobs` above.

psarando (approved):
LGTM 👍 🎉 ✨

ianmcorvidae:
Thanks for the review! I left a comment on that Jira issue -- it didn't seem immediately related to me, so I didn't close it outright, but thought I should note it here at least. I'll merge this now despite the release tomorrow, since it fixes a bug and otherwise shouldn't be too huge a change, but I'll at least try to test it a little in QA tonight too.

PR #266, opened 2024-01-03
CORE-1971: manage default memory & CPU limits as configuration values in the apps service instead of hardcoded values in app-exposer

ianmcorvidae:
Hopefully fairly easily understood. For the UI, we've added an include-defaults parameter (cyverse-de/common-swagger-api#84 and cyverse-de/terrain#277 for the schema and ensuring that reaches this service, resp.) so that when _editing_ tools the configured default isn't listed as though it's the tool-specific configured value. When launching apps, however, the default should show up seamlessly as though it's configured in the tool & passed down to app-exposer.

This still needs testing, but as discussed elsewhere I might do so in the QA environment to ease the cross-service setup process.

### DIFF
--- a/src/apps/service/apps/de/listings.clj
+++ b/src/apps/service/apps/de/listings.clj
@@ -687,7 +687,7 @@
 (defn- format-task
   [user {:keys [tool_id] :as task}]
   (-> task
-      (assoc :tool (when tool_id (tools/get-tool user tool_id)))
+      (assoc :tool (when tool_id (tools/get-tool user tool_id true)))
       (dissoc :tool_id)
       (update-in [:inputs] (partial map (comp remove-nil-vals format-task-file-param)))
       (update-in [:outputs] (partial map (comp remove-nil-vals format-task-output)))
--- a/src/apps/tools.clj
+++ b/src/apps/tools.clj
@@ -88,10 +88,11 @@
 
 (defn get-tool
   "Obtains a tool by ID."
-  [user tool-id]
+  [user tool-id include-defaults]
   (let [tool           (->> (persistence/get-tool tool-id)
                             (format-tool-listing (perms-client/load-tool-permissions user)
                                                  (perms-client/get-public-tool-ids)))
+        ;; XXX: use include-defaults here
         container      (clojure.set/rename-keys (tool-container-info tool-id) {:ports :container_ports})
         implementation (persistence/get-tool-implementation-details tool-id)]
     (assoc tool
@@ -100,9 +101,9 @@
 
 (defn user-get-tool
   "Obtains tool details for a user."
-  [user tool-id]
+  [user tool-id include-defaults]
   (permissions/check-tool-permissions user "read" [tool-id])
-  (get-tool user tool-id))
+  (get-tool user tool-id include-defaults))
 
 (defn- add-new-tool
   [{:keys [container] :as tool}]
@@ -136,7 +137,7 @@
    (persistence/update-tool tool)
    (when container
      (set-tool-container id overwrite-public container))
-   (get-tool user id)))
+   (get-tool user id false)))
 
 (defn delete-tool
   [tool-id]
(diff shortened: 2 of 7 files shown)


src/apps/service/apps/de/listings.clj
```diff
@@ -687,7 +687,7 @@
 (defn- format-task
   [user {:keys [tool_id] :as task}]
   (-> task
-      (assoc :tool (when tool_id (tools/get-tool user tool_id)))
+      (assoc :tool (when tool_id (tools/get-tool user tool_id true)))
```
ianmcorvidae, line 690: I _think_ this is all that's needed to get the default values passed down to app-exposer/jex-adapter, but it feels a bit too easy so I feel like I must have missed something, haha.

src/apps/tools.clj
```diff
@@ -54,6 +54,14 @@
               :tool_request_status)
       remove-nil-vals))
 
+(defn- format-container-settings
+  [container-settings include-defaults]
+  (if include-defaults
+      (merge {:max_cpu_cores (config/default-cpu-limit)
+              :memory_limit (config/default-memory-limit)}
+             container-settings)
```
ianmcorvidae, line 62: `merge` overwrites values from the first map with ones from the second, so this should suffice, I think

src/apps/tools.clj
```diff
@@ -136,7 +144,7 @@
    (persistence/update-tool tool)
    (when container
      (set-tool-container id overwrite-public container))
-   (get-tool user id)))
+   (get-tool user id false)))
```
ianmcorvidae, line 147: I figured this and other 'update'/'set' type functions shouldn't show the defaults.

src/apps/util/config.clj
```diff
@@ -123,6 +123,16 @@
   [props config-valid configs]
   "apps.tools.private.memory-limit" (* 16 1024 1024 1024)) ;; 16GB
 
+(cc/defprop-optint default-cpu-limit
+  "The CPU limit, in cores (matching the database's data type), to be used as the default resource limit when one is not set in a tool definition."
+  [props config-valid configs]
+  "apps.tools.default.cpu-limit" 4)
```
ianmcorvidae, line 129: Unsure if this is too low, but I just copied it from the app-exposer hardcoded default for now.

src/apps/util/config.clj
```diff
@@ -123,6 +123,16 @@
...
+
+(cc/defprop-optint default-memory-limit
+  "The memory limit, in bytes, to be used as the default resource limit when one is not set in a tool definition."
+  [props config-valid configs]
+  "apps.tools.default.memory-limit" (* 16 1024 1024 1024)) ;; 16GB
```
ianmcorvidae, line 134: This one is higher than the hardcoded one in app-exposer, which is 8Gi instead, in order to match the limit when adding private tools that's further up in this file. Plus, 8Gi is pretty small for a limit anyway.

ianmcorvidae (commented):
(some miscellaneous comments/thoughts)

src/apps/tools.clj
slr71: I think that should suffice as well. 👍

src/apps/service/apps/de/listings.clj
slr71: I feel that way a lot when I'm doing Clojure programming. 😆

src/apps/tools.clj
slr71: That sounds reasonable to me too.

src/apps/util/config.clj
slr71: That seems like a reasonable default. We can always tweak it in the configuration files if necessary.

src/apps/util/config.clj
slr71: 👍

slr71 (approved):
Looks good to me!

PR #268, opened 2024-03-15
Only set the default maximums for interactive jobs

ianmcorvidae:
This is only for the spot to do with launching jobs. The tool editing bits should, I think, already be doing the right thing. Or right enough, at least.


slr71:
Looks good to me! 👍

PR #270, opened 2024-05-31
CORE-1889 Tapis v3 Migration

psarando:
This PR will rename everything from `agave` to `tapis` and add updates for cyverse-de/mescal#25 and cyverse-de/common-swagger-api#85, which must be merged first before this PR can be merged.


src/apps/persistence/oauth.clj
```diff
@@ -73,8 +66,7 @@
 
 (defn store-access-token
   "Stores information about an OAuth access token in the database."
-  [api-name username {:keys [token-type expires-at refresh-token access-token]}]
-  (validate-token-type token-type)
```
psarando, line 77: Tapis no longer returns a `token-type` field.

src/apps/service/apps/combined.clj
```diff
@@ -365,4 +365,5 @@
     (.hasAppPermission (util/get-apps-client clients system-id) username system-id app-id required-level))
 
   (supportsJobSharing [_ job-step]
-    (.supportsJobSharing (util/apps-client-for-job-step clients job-step) job-step)))
+    (and (not= (:job_type job-step) jp/agave-job-type)
+         (.supportsJobSharing (util/apps-client-for-job-step clients job-step) job-step))))
```
psarando, line 369: The changes in this namespace are only for allowing users to still list their old Agave jobs, even if they can't relaunch them anymore.

Maybe it would be better to re-add an `Agave` type that handles the logic in these methods, and throws an error for the rest? 🤷

src/apps/service/apps/jobs/util.clj
```diff
@@ -51,8 +52,7 @@
    "FolderInput"       "collection"
    "MultiFileSelector" "many"})
 
-(def input-types
-  (set (keys input-multiplicities)))
+(def input-types am/param-ds-input-types)
```
psarando, line 55: The HPC `FileFolderInput` type was missing from this before, so inputs were not repopulating in the UI relaunch forms correctly.

src/apps/service/callbacks.clj
```diff
@@ -11,10 +11,13 @@
...
-  [job-id last-updated {:keys [status external-id end-time]}]
+(defn update-tapis-job-status
+  [job-id {{:keys [timestamp type data]} :event}]
   (service/assert-valid job-id "no job UUID provided")
-  (service/assert-valid status "no status provided")
-  (service/assert-valid external-id "no external job ID provided")
-  (log/info (str "received a status update for Agave job " external-id ": status = " status))
-  (apps/update-job-status job-id external-id status (first (remove string/blank? [end-time last-updated]))))
+  (service/assert-valid type "no status provided")
+  (let [{:keys [jobUuid]} (service/parse-json data)
```
psarando, line 18: Not sure why once again, but this `data` field is JSON encoded as a string, so we have to parse it here in order to get the Tapis job ID.

src/apps/service/apps/combined.clj
slr71: That seems pretty clean to me, but this is also understandable, so I'm open to leaving it as is.

slr71 (approved):
Looks good to me! :shipit:

psarando:
Thanks for the review! 👍

PR #271, opened 2024-07-25
CORE-2006: update dependencies.

slr71:
This is one of the more complex dependency removal updates and several overarching changes were made:

- Removed support for ping events over AMQP, which haven't been used since we switched to Kubernetes.
- The clostache dependency was throwing a name redefinition error, so I replaced it with a homegrown function. The new function doesn't have nearly as many features as Clostache, but it handles all of the cases that we currently need.
- Renamed some functions to get rid of more name redefinition errors.
- Removed calls to compojure.route/middleware to fix some deprecation warnings.
- I encountered some errors related to opentelemetry when I was doing testing. We haven't been using opentelemetry lately, so instead of troubleshooting the problem, I simply removed support for it.
- Removed AMQP support entirely because it turned out that the only reason we needed AMQP was for the pings that we don't use anymore.
- Switched to a new Docker base image.

My biggest concerns at the moment are the changes that I made to remove calls to `compojure.route/middleware` and to remove support for opentelemetry. If extra scrutiny can be given to those changes, it would be helpful.


src/apps/service/apps/combined/util.clj
```diff
@@ -16,7 +16,7 @@
   "Expects results to be a list of maps in a format like {:total int, :apps []}"
   [params results]
   (let [params      (apply-default-search-params params)
-        maybe-deref (fn [r] (if (future? r) @r r))
```
slr71, line 19: When I was testing, I encountered an error in this function because `r` was `nil` one time. I haven't been able to duplicate it since. This should fix the problem, but I can't be sure since I haven't been able to reproduce the problem.

psarando (changes requested):
I think I found a bug, otherwise LGTM 👍

src/apps/service/apps/combined/util.clj
```diff
@@ -1,5 +1,5 @@
 (ns apps.service.apps.combined.util
-  (:use [apps.service.util :only [sort-apps apply-offset apply-limit uuid?]]
+  (:use [apps.service.util :only [sort-apps apply-offset apply-limit valid-uuid?]]
```
psarando, line 2: I don't think `valid-uuid?` is used anywhere in this file.

src/apps/clients/async_tasks.clj
```diff
@@ -35,13 +34,8 @@
...
   [async-task-id thread-function prefix]
-  (otel/with-span [outer-span ["run-async-thread" {:kind :producer :attributes {"async-task-id" (str async-task-id)}}]]
-    (let [^Runnable task-thread (fn []
-                                  (with-open [_ (otel/span-scope outer-span)]
-                                    (otel/with-span [s ["async thread" {:kind :consumer :attributes {"async-task-id" (str async-task-id)}}]]
-                                      (thread-function async-task-id))))]
-      (.start (Thread. task-thread (str prefix "-" (string/replace async-task-id #".*/tasks/" "")))))
-    async-task-id))
+  (let [^Runnable task-thread (fn [] (thread-function async-task-id))]
+    (.start (Thread. task-thread (str prefix "-" (string/replace async-task-id #".*/tasks/" ""))))))
```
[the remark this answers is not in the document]
slr71: Ah, good catch!

src/apps/service/util.clj
```diff
@@ -54,18 +53,18 @@
...
+(defn valid-uuid?
   [s]
-  (or (instance? UUID s)
+  (or (uuid? s)
       (re-find #"\A\p{XDigit}{8}(?:-\p{XDigit}{4}){3}-\p{XDigit}{12}\z" s)))
 
 (defn extract-uuids
   [ids]
-  (seq (map uuids/uuidify (filter uuid? ids))))
+  (seq (map uuids/uuidify (filter valid-uuid? ids))))
```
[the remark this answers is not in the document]
slr71: Good catch! I'll remove this too.

src/apps/service/apps/combined/util.clj
slr71: I'll remove that too. Thanks!

psarando (approved):
LGTM 👍

slr71:
Thanks for the review! 👍

PR #272, opened 2024-07-26
Handle username suffix better

ianmcorvidae:
This tries to centralize the logic for username suffix handling to all use a regex-type approach like I used in the analyses service (remove everything starting from the first `@`, then add the suffix after that). This moves most places that cared about the configuration property to use the same function within apps.user, with one exception that was set up to pass the uid domain in. That could maybe go away too, but thought I'd get this much up at least.


slr71 (approved):
Looks good to me! 👍

psarando (approved):
LGTM 👍

PR #274, opened 2024-09-25
Route periodic notification information into the stored job submission

ianmcorvidae:
Also includes the start of the update for cyverse-de/common-swagger-api#87 which can be updated to a non-snapshot version when that's in properly.


ianmcorvidae:
Okay, upon testing this actually isn't even necessary, because this codepath isn't followed before storing the submission in the DB, only for what gets passed down to jex-adapter/app-exposer. Closing this since the change isn't needed.

PR #275, opened 2024-10-10
CORE-2011 Update max CPU reconciliation in DE job submissions.

psarando:
Updated `apps.service.apps.de.jobs.common/reconcile-container-requirements` to use `0` for the max CPU request for user submissions if nothing was submitted from the client.

This will prevent the maximum CPUs set by the tool from being used in the final job submission, allowing the job services to use their own reasonable default instead (currently `4`).

A similar check was added for user `memory_limit` requests, which will also allow job services to use their own defaults (VICE jobs will use `8Gi`, for example).

### DIFF
--- a/src/apps/service/apps/de/jobs/common.clj
+++ b/src/apps/service/apps/de/jobs/common.clj
@@ -64,6 +64,9 @@
                           :max_cpu_cores]))
 
 (defn- filter-request-max-requirements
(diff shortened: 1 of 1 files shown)


ianmcorvidae (approved):
looks reasonable/straightforward to me, at least!

psarando:
Thanks for the review!

PR #276, opened 2024-10-16
CORE-1955 Add endpoint for reordering app versions

psarando:
Added `PUT /apps/:system/:id/versions` for reordering app versions.

This endpoint expects the app versions to be in descending order, with the newest (or latest) first, since that is also the way they are listed in the app details and listing endpoints.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -30,7 +30,7 @@
                  [org.cyverse/metadata-client "3.1.2"]
                  [org.cyverse/common-cli "2.8.2"]
                  [org.cyverse/common-cfg "2.8.3"]
-                 [org.cyverse/common-swagger-api "3.4.6"]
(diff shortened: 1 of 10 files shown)


ianmcorvidae (approved):
I didn't spot anything, seems good to me.

psarando:
I overlooked the fact that `POST /apps/:system/:id/versions` already exists for adding a new app version, so I'll update this endpoint for reordering app versions to `PUT /apps/:system/:id/versions`.

ianmcorvidae:
Ah, that makes sense. PUT is probably more appropriate for this anyway since it's a replace rather than an add, so sounds good to me.

slr71 (approved):
I didn't spot any problems. Looks good to me! :shipit:

psarando:
Thanks for the reviews! 👍

PR #279, opened 2024-12-14
CORE-2024: memory and cpu request limit fixes

slr71:
We were encountering a problem where the memory requests weren't being honored for VICE job submissions. This is an attempt to fix the problem.

### DIFF
--- a/src/apps/service/apps/de/jobs/common.clj
+++ b/src/apps/service/apps/de/jobs/common.clj
@@ -56,11 +56,12 @@
 (defn- reconcile-container-requirements
   "reconcile submission requirement requests with tool requirements"
   [container requirements]
-  (assoc container
-         :min_memory_limit (resources/get-required-memory container requirements)
-         :memory_limit     (resources/get-max-memory container requirements)
-         :min_cpu_cores    (resources/get-required-cpus container requirements)
-         :max_cpu_cores    (resources/get-max-cpus container requirements)))
+  (-> (assoc container
+             :min_memory_limit (resources/get-required-memory container requirements)
+             :memory_limit     (resources/get-max-memory container requirements)
+             :min_cpu_cores    (resources/get-required-cpus container requirements)
+             :max_cpu_cores    (resources/get-max-cpus container requirements))
+      remove-nil-vals))
 
 (defn- add-container-info
   [{tool-id :id tool-type :type :as component} requirements]
--- a/src/apps/service/apps/de/jobs/common.clj
+++ b/src/apps/service/apps/de/jobs/common.clj
@@ -9,6 +9,7 @@
             [apps.containers :as c]
             [apps.tools :as t]
             [apps.service.apps.de.jobs.params :as params]
+            [apps.service.apps.de.jobs.resources :as resources]
             [apps.service.apps.jobs.util :as util]
             [clojure.string :as string]))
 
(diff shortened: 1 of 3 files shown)


src/apps/service/apps/de/jobs/common.clj
```diff
@@ -52,45 +53,14 @@
...
-             (merge-with min
-                         (filter-container-max-requirements container)
-                         (filter-request-max-requirements requirements)))
-      (limit-container-min-requirement :min_memory_limit :memory_limit)
-      (limit-container-min-requirement :min_cpu_cores :max_cpu_cores)))
+  (assoc container
+         :min_memory_limit (resources/get-required-memory container requirements)
+         :memory_limit     (resources/get-max-memory container requirements)
+         :min_cpu_cores    (resources/get-required-cpus container requirements)
+         :max_cpu_cores    (resources/get-max-cpus container requirements)))
```
slr71, line 63: This feels a lot less elegant than the code that was there before, but I wanted to make things explicit in order to make it easier to reason about the requests and limits.

src/apps/service/apps/de/jobs/common.clj
```diff
@@ -103,8 +73,9 @@
      component)
    :id))
 
-(defn- load-step-component
-  [task-id requirements]
+
+(defn- load-tool-info
+  [task-id]
```
slr71, line 78: I split `load-tool-info` out from `load-step-component` for debugging, and I decided just to leave it that way.

src/apps/service/apps/de/jobs/resources.clj
```diff
@@ -0,0 +1,56 @@
...
+  (let [c-max (get container max-kw)
+        r-max (get requirements max-kw)]
+    (cond
+      ;; If the requested maximum wasn't specified, use the requested minimum.
+      (or (nil? r-max) (zero? r-max) (neg? r-max))
+      (min-resource-setting-fn container requirements)
+
+      ;; If the requested maximum was specified, it must not be greater than the container maximum.
+      :else
+      (min (or c-max default) r-max))))
```
slr71, line 36: This is where the bulk of the logic is for this project, so this is where I'd like the most scrutiny.

test/apps/service/apps/resources_test.clj
```diff
@@ -0,0 +1,78 @@
...
+(def gib (* 1024 1024 1024))
+
+(defn reqs
+  [min-mem max-mem min-cpu max-cpu]
+  {:min_memory_limit (when min-mem (* min-mem gib))
+   :memory_limit     (when max-mem (* max-mem gib))
+   :min_cpu_cores    min-cpu
+   :max_cpu_cores    max-cpu})
+
+(def test-cases
```
psarando, line 14: Good test cases! 🎉

test/apps/service/apps/resources_test.clj
```diff
@@ -0,0 +1,78 @@
...
+    :expected     (reqs 4 4 4 4)
+    :desc         "maximum requests not specified"}
+   {:container    (reqs 2 256 2 256)
+    :requirements (reqs nil nil nil nil)
+    :expected     (reqs 2 2 2 2)
+    :desc         "no requests specified"}
+   {:container    (reqs nil nil nil nil)
+    :requirements (reqs 32 32 32 32)
+    :expected     (reqs 16 16 4 4)
+    :desc         "no cpu settings specified"}])
```
psarando, line 70: Should this `desc` be "no container settings specified"?

test/apps/service/apps/resources_test.clj
```diff
@@ -0,0 +1,78 @@
...
+    :expected     (reqs 4 256 4 256)
+    :desc         "maximum requests greater than container maximums"}
+   {:container    (reqs 2 256 2 256)
+    :requirements (reqs 4 nil 4 nil)
+    :expected     (reqs 4 4 4 4)
+    :desc         "maximum requests not specified"}
+   {:container    (reqs 2 256 2 256)
+    :requirements (reqs nil nil nil nil)
+    :expected     (reqs 2 2 2 2)
+    :desc         "no requests specified"}
```
psarando, line 66: Should we add a test case for the issue that #275 intended to fix?

If the container has no minimum requirement, and the client does not submit anything for that min or max requirement, then this service was supposed to submit `0` (or `nil`?) to the jobs services, to allow them to use their own defaults for that requirement.

The default can be different for VICE and condor jobs. For example, VICE will use `8GiB` for its default max memory when nothing is requested, but condor jobs will not submit any value for max memory (allowing the job to use all available memory on the node?). They might be using the same default of `4` for max CPUs.

```clojure
   {:container    (reqs nil 256 nil 256)
    :requirements (reqs nil nil nil nil)
    :expected     (reqs 0 0 0 0)
    :desc         "no requests specified, and no container minimums specified"}
```

Or do we now prefer the apps service to enforce a default max if nothing is requested from the client?

psarando (commented):
I think I reproduced the issue described by CORE-2024, so I see now that #275 appears to have introduced the bug that this PR will fix:

I submitted a VICE job request for an app where its tool had a `memory_limit` of 32GiB, and my submission requested 32GiB for the `min_memory_limit` requirement, but it seems this service submitted 0 to the VICE jobs service, which used its own default of 8GiB when the job ran.

test/apps/service/apps/resources_test.clj
slr71: Ah, that's a good question. I should probably support this test case as well. I'll take a look at this a little later.

test/apps/service/apps/resources_test.clj
slr71: Ah! I knew I'd miss one of these copy pasta errors. 😆

slr71:
I'm going to merge this now so that we can get the changes deployed. Thanks for the review! More reviews are still welcome as well.

PR #285, opened 2025-06-28
CORE-2064: fix an error in the app launch view endpoint.

slr71:
The purpose of this PR is to fix an error that was calling whenever the `GET /apps/de/{app-id}` endpoint was called in the apps service. This is the error that was occurring:

```
org.postgresql.util.PSQLException: Can't infer the SQL type to use for an instance of korma.sql.fns$pred_not. Use setObject() with an explicit Types value to specify the type to use.
```

This was caused by a bad translation of some code that used Korma from using the symbols, `not` and `exists` to using the corresponding keywords. I would have had to explicitly indicate that `:exists` should be treated like an SQL function and I didn't do that.

Korma hasn't been maintained for several years, so I decided to switch over to HoneySQL instead, which doesn't require the use of macros and is actively maintained.


project.clj
```diff
@@ -23,15 +23,15 @@
                  [metosin/compojure-api "1.1.14"]
                  [org.cyverse/async-tasks-client "0.0.5"]
                  [org.cyverse/authy "3.0.1"]
-                 [org.cyverse/clj-kondo-exports "0.1.0-SNAPSHOT"]
+                 [org.cyverse/clj-kondo-exports "0.1.0"]
```
slr71, line 26: Not sure why I failed to push this commit to the main fork before, but I'm glad that this was caught before next week's maintenance.

skaffold.yaml
```diff
@@ -1,14 +1,17 @@
-apiVersion: skaffold/v1
+apiVersion: skaffold/v3
```
slr71, line 1: `skaffold/v1` doesn't support platform specifications, which prevents me from running `skaffold dev` or `skaffold run` on my ARM based Macs.

src/apps/service/apps/de/job_view.clj
```diff
@@ -33,28 +37,31 @@
...
+                [:= :iom.input :p.id]
+                [:= :wm.target_step step-id]])))
+
+(defn- get-parameters-query
+  [step-id group-id include-hidden-params?]
+  (-> (mp/hsql-params-base-query)
+      (h/order-by :p.display-order)
+      (h/where [:and
+                [:= :p.parameter_group_id group-id]
+                (when include-hidden-params? :p.is_visible)
```
slr71, line 53: I just learned about this feature of HoneySQL today. The `:and` and `:or` operators ignore nil clauses by default, so code like this no longer requires helper functions each time we want to conditionally add a clause.

src/apps/service/apps/de/job_view.clj
```diff
@@ -33,28 +37,31 @@
...
+                [:and
+                 [:not [:exists (mapped-input-subselect step-id)]]
+                 [:or
+                  [:= :p.value_type "Input"]
+                  [:not [:coalesce :is_implicit false]]]]])
```
slr71, line 58: I decided to switch from having two separate clauses to check for null and false values to using `coalesce` instead. It seemed a little cleaner to me.

PR #295, opened 2026-04-25
CORE-2139: ensure that jobs for which downstream submissions fail are…

slr71:
… still recorded in the database

This replaces the higher level transactions in `apps.services.apps.jobs.submissions.submit` and `apps.services.apps.jobs.submissions.async` with lower level transactions in order to ensure that the rows are recorded in the database if a submission failure occurs.


psarando (commented):
I didn't catch any issues 👍

slr71:
Thanks for the review! 👍

PR #297, opened 2026-05-14
Build interactive analysis URLs from per-operator base URL

johnworth:
## Summary

- The job base queries (`job-base-query`, `hsql-job-base-query`) select `operator_base_url` straight from the `job_listings` view, which now left-joins `operators` itself.
- `interactive-urls` builds the URL from `operator_base_url`, falling back to the static `interapps-base` config for jobs with no operator (legacy / pre-operator launches).
- `send-interactive-job-status-update` reuses the formatted job's `:interactive_urls` instead of rebuilding the URL from the static base, so notification/email links match the listing.

## Context

VICE analyses can now be scheduled onto different Kubernetes clusters, each fronted by its own `vice-operator` with its own landing domain. Previously `apps` built every interactive URL from a single static config value, so an analysis launched on a non-default cluster linked the user to the wrong domain — an infinitely-refreshing loading page on a cluster that wasn't running the analysis.

Depends on cyverse-de/de-database#37 (adds `operators.base_url` and exposes `operator_base_url` in the `job_listings` view) and pairs with cyverse-de/app-exposer#145. All three repos share the `operator-base-url` branch.

## Test plan

- [x] `lein check`, `lein test` (71 tests, 0 failures)
- [ ] launch a VICE analysis on a non-default operator; confirm the job listing's `interactive_urls` and the status-update notification both point at that operator's domain
- [ ] a job with no operator still resolves via `interapps-base`

### DIFF
--- a/src/apps/clients/notifications.clj
+++ b/src/apps/clients/notifications.clj
@@ -106,11 +106,14 @@
    (send-job-status-update username email-address job-info)))
 
 (defn send-interactive-job-status-update
-  "Sends notification of an interactive job status update to the user."
-  ([username email-address {status :status user-id :user_id :as job-info} {external-id :external_id}]
-   (if-not (jp/completed? status)
-     (let [access-url (str (interapps-url (curl/url (config/interapps-base)) user-id external-id))]
-       (send-job-status-update username email-address (assoc job-info :access_url access-url)))
+  "Sends notification of an interactive job status update to the user. The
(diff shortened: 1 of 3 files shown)


johnworth:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #298, opened 2026-06-10
Add apps.util.cache with some simple caches 

ianmcorvidae:
(for immutable or rarely changed data that still gets looked up a lot)

Right now just used for get-workspace and get-active-hierarchy-version, both of which get called repeatedly in some places but are largely immutable. Found by way of looking for admin endpoint performance gains.


psarando (approved):
LGTM 👍

ianmcorvidae:
Thanks for the review!

PR #299, opened 2026-06-11
Fix N+1 query pattern in list-app-publication-requests

ianmcorvidae:
Refactored the publication requests endpoint to batch all I/O instead of calling get-app-details* per request

- Added batch DB functions: get-app-latest-versions, get-app-version-details-batch, get-app-version-tools-batch, get-app-references-for-versions, get-version-ids-with-docs, get-groups-for-apps, get-suggested-groups-for-apps, get-public-job-stats-batch
- Refactored format-app-details to accept optional pre-fetched data maps, so batched and single-app paths share formatting logic
- Perform independent I/O concurrently via futures

Depends on https://github.com/cyverse-de/metadata-client/pull/8 and transitively https://github.com/cyverse-de/metadata/pull/30 and https://github.com/cyverse-de/common-swagger-api/pull/99

### DIFF
--- a/src/apps/clients/metadata.clj
+++ b/src/apps/clients/metadata.clj
@@ -67,7 +67,6 @@
   [username app-ids attrs]
   (let [ontology-version (get-active-hierarchy-version :validate false)]
     (if (and ontology-version (seq app-ids))
-      #_:clj-kondo/ignore
       (let [result (metadata-client/filter-hierarchies-batch
                     (config/metadata-client)
                     username
--- a/src/apps/clients/metadata.clj
+++ b/src/apps/clients/metadata.clj
@@ -67,6 +67,7 @@
   [username app-ids attrs]
   (let [ontology-version (get-active-hierarchy-version :validate false)]
     (if (and ontology-version (seq app-ids))
+      #_:clj-kondo/ignore
       (let [result (metadata-client/filter-hierarchies-batch
                     (config/metadata-client)
                     username
--- a/src/apps/clients/metadata.clj
+++ b/src/apps/clients/metadata.clj
@@ -1,6 +1,7 @@
 (ns apps.clients.metadata
   (:require [apps.persistence.categories :as db-categories]
             [apps.util.config :as config]
+            [kameleon.uuids :refer [uuidify]]
             [metadata-client.core :as metadata-client]
             [slingshot.slingshot :refer [throw+]]))
 
@@ -59,6 +60,25 @@
                                       attrs
                                       app-target-type
                                       app-id))
+
+(defn filter-hierarchies-batch
+  "Fetches ontology hierarchy classifications for multiple apps in one request.
+   Returns a map of app-id -> {:hierarchies [...]}."
+  [username app-ids attrs]
+  (let [ontology-version (get-active-hierarchy-version :validate false)]
+    (if (and ontology-version (seq app-ids))
+      (let [result (metadata-client/filter-hierarchies-batch
+                    (config/metadata-client)
+                    username
+                    ontology-version
+                    attrs
+                    [app-target-type]
+                    app-ids)]
+        ;; result is {:targets [{:id UUID :hierarchies [...]} ...]}
+        (into {} (map (fn [{:keys [id hierarchies]}]
+                        [(uuidify id) {:hierarchies hierarchies}]))
+              (:targets result)))
+      {})))
 
 (defn filter-targets-by-ontology-search
   [username category-attrs search-term app-ids & {:keys [validate] :or {validate true}}]
--- a/src/apps/clients/permissions.clj
+++ b/src/apps/clients/permissions.clj
@@ -204,7 +204,6 @@
 (defn- get-resource-users
   "Lists users with any level of access to a resource."
   [resource-type resource-name]
-  #_:clj-kondo/ignore
   (->> (pc/list-resource-permissions (client) resource-type resource-name true)
        :permissions
        (mapv (comp :subject_id :subject))))
--- a/src/apps/clients/permissions.clj
+++ b/src/apps/clients/permissions.clj
@@ -204,6 +204,7 @@
 (defn- get-resource-users
   "Lists users with any level of access to a resource."
   [resource-type resource-name]
+  #_:clj-kondo/ignore
   (->> (pc/list-resource-permissions (client) resource-type resource-name true)
        :permissions
        (mapv (comp :subject_id :subject))))
--- a/src/apps/persistence/app_documentation.clj
+++ b/src/apps/persistence/app_documentation.clj
@@ -1,5 +1,6 @@
 (ns apps.persistence.app-documentation
   (:require [apps.persistence.entities :refer [app_references]]
+            [apps.util.db :as db]
             [korma.core :as sql]))
 
 (defn get-app-references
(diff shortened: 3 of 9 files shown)


src/apps/service/apps/de/listings.clj
```diff
@@ -768,17 +803,80 @@
...
-  "Lists app publication requests, optionally filtering the listing by app ID or requestor."
+  "Lists app publication requests, optionally filtering the listing by app ID or requestor.
+   Uses batched queries to avoid the N+1 problem."
   [user {app-id :app_id include-completed :include_completed :keys [requestor]}]
-  (mapv (partial format-app-publication-request user)
-        (amp/list-app-publication-requests app-id requestor include-completed)))
+  (let [requests (amp/list-app-publication-requests app-id requestor include-completed)]
+    (if (empty? requests)
+      []
+      (let [{:keys [username shortUsername]} user
```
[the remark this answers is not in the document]
ianmcorvidae: There's no real test harness we can tie into here, which would make this a huge change unfortunately.

src/apps/service/apps/de/listings.clj
```diff
@@ -587,41 +597,66 @@
...
+                                        (get groups-map app-id [])
+                                        (get-groups-for-app app-id))
+                :suggested_categories (if suggested-map
+                                        (get suggested-map app-id [])
+                                        (get-suggested-groups-for-app app-id))
+                :system_id            c/system-id)
+         ((fn [app]
+            (if hierarchies-map
+              (merge app (get hierarchies-map app-id {:hierarchies []}))
+              (format-app-hierarchies app username))))
```
psarando: I think if this were refactored to use `as->`, then this wouldn't have to be wrapped in a `fn`?
This works just as well, though.

src/apps/clients/metadata.clj
```diff
@@ -60,6 +61,26 @@
                                       app-target-type
                                       app-id))
 
+(defn filter-hierarchies-batch
+  "Fetches ontology hierarchy classifications for multiple apps in one request.
+   Returns a map of app-id -> {:hierarchies [...]}."
+  [username app-ids attrs]
+  (let [ontology-version (get-active-hierarchy-version :validate false)]
+    (if (and ontology-version (seq app-ids))
+      #_:clj-kondo/ignore
```
slr71: I wasn't sure which lint warning was being ignored here. I tried removing the annotation, and running a lint check. Nothing came up, I might have a configuration setting that's preventing it from showing up, though.

I did see a "redundant ignore" warning when I ran the linter with the annotations in place. I'd be tempted to remove these and see what happens.

src/apps/clients/permissions.clj
```diff
@@ -204,6 +204,7 @@
 (defn- get-resource-users
   "Lists users with any level of access to a resource."
   [resource-type resource-name]
+  #_:clj-kondo/ignore
```
slr71: I'd be tempted to remove this as well and see if any warnings are still being reported.

src/apps/persistence/app_documentation.clj
```diff
@@ -3,10 +3,18 @@
...
   [app-version-id]
   (sql/select app_references (sql/where {:app_version_id app-version-id})))
 
+(defn get-app-references-for-versions
+  "Retrieves references for multiple app version IDs.
+   Returns a map of version-id -> [reference-row ...]."
+  [version-ids]
+  (when (seq version-ids)
+    (group-by :app_version_id
+              (sql/select app_references (sql/where {:app_version_id [:in version-ids]})))))
```
slr71: This probably doesn't apply here, but if the list is extremely long (that is, contains more than about 32k items) then it might be necessary to use `apps.util.db/sql-array` to avoid errors in queries like this one.

src/apps/service/apps/de/listings.clj
```diff
@@ -768,17 +803,80 @@
...
-  {:id        id
-   :app       (get-app-details* user app-id false)
-   :requestor requestor})
+(defn- format-app-publication-request-batched
+  "Formats a single publication request using pre-fetched batch data.
+   Delegates to format-app-details with pre-fetched maps to avoid duplicating formatting logic."
+  [user format-app details-map tools-map refs-map prefetched
+   {app-id :app_id :keys [id requestor]}]
+  (let [details (get details-map app-id)
+        version-id (:version_id details)
```
slr71, line 812: This is a nitpick, but other variables and expressions are aligned in this set of bindings, but this one isn't.

slr71 (approved):
Looks good to me! I had a couple of nitpicks, but I didn't spot anything that would cause a problem. 👍

src/apps/clients/metadata.clj
ianmcorvidae: Interesting. I was getting a complaint about unknown symbol, despite it being present in the metadata client. Maybe I'll check if I'm running an old version of clj-kondo or something. Or just remove the ignores, but it was blocking my pre-commit hooks unless I threw on `--no-verify`, heh

src/apps/service/apps/de/listings.clj
ianmcorvidae: No sense not making the machine align the text! Thanks for spotting it

src/apps/persistence/app_documentation.clj
ianmcorvidae: That's a good point. I'll try to get that refactored to use that, since doesn't seem like there's much point in leaving it the more brittle version anyway.

src/apps/persistence/app_listing.clj
```diff
@@ -414,6 +414,33 @@
...
+    (let [rating-avg-subselect (sql/subselect
+                                entities/ratings
+                                (sql/fields (sql/raw "CAST(COALESCE(AVG(rating), 0.0) AS DOUBLE PRECISION) AS average_rating"))
+                                (sql/where {:ratings.app_id
+                                            :app_versions_listing.id}))
+          rating-total-subselect (sql/subselect
+                                  entities/ratings
+                                  (sql/aggregate (count :ratings.rating) :total_ratings)
+                                  (sql/where {:ratings.app_id
+                                              :app_versions_listing.id}))
```
psarando: These subselects appear to be exact duplicates of the subselects used in `get-app-version-details` above. I'd probably refactor them into a common place.

src/apps/persistence/app_metadata.clj
```diff
@@ -1125,3 +1125,47 @@
...
+  "Retrieves the list of available versions for multiple apps.
+   Returns a map of app-id -> [{:version ... :version_id ...} ...]."
+  ([app-ids]
+   (list-app-versions-for-apps app-ids false))
+  ([app-ids include-deleted?]
+   (when (seq app-ids)
+     (let [rows (-> (select* entities/app_versions)
+                    (fields :app_id :version [:id :version_id])
+                    (where {:app_id [:in app-ids]})
+                    ((fn [q] (if-not include-deleted? (where q {:deleted false}) q)))
```
psarando: There's already an `add-deleted-versions-clause` function in this namespace.

psarando (approved):
I just had a couple of nits to pick, but these looks like nice improvements!
LGTM 👍 🎉

src/apps/service/apps/de/listings.clj
ianmcorvidae: That's a great idea, yeah. Updated to do that.

src/apps/persistence/app_listing.clj
ianmcorvidae: Refactored them out, and asked the machine to find anything else it could to restructure a bit for reuse too. I think the result is better overall, at least.

ianmcorvidae:
I'm going to go ahead and merge this now and hopefully get it all deployed so I can test that it actually works. Any additional reviews are welcome of course!

psarando (commented):
LGTM 🎉 👍

PR #300, opened 2026-06-15
Don't uuidify session IDs to correspond to de-database migration 52

ianmcorvidae:
goes with https://github.com/cyverse-de/de-database/pull/41


ianmcorvidae:
I'll merge this as soon as the database migration is in place, but reviews will still be welcome after that.

PR #301, opened 2026-07-21
Forward mount_data_store from analysis submission into job model

ianmcorvidae:
The field existed in the AnalysisSubmission schema and in model.Job, but build-submission silently dropped it. Add it to the outgoing job map with a default of true so that omitting the field from a submission preserves the existing behaviour.


psarando (approved):
LGTM 👍

psarando (approved):
Also LGTM 👍

ianmcorvidae:
I'm going to go ahead and merge this and get it deployed, along with the terrain update to common-swagger-api. Neither should do anything until the sonora or app-exposer PRs are merged since it's basically just plumbing.

PR #302, opened 2026-08-03
Pass through initial time limits and maximums, and set maximum value in config

ianmcorvidae:


src/apps/service/apps/jobs/submissions.clj
```diff
@@ -216,6 +216,14 @@
...
 (defn submit
   [apps-client user {app-id :app_id system-id :system_id :as submission}]
+  (when-let [tls (:time_limit_seconds submission)]
+    (when-not (pos? tls)
+      (throw+ {:type  :clojure-commons.exception/bad-request
+               :error "time_limit_seconds must be a positive integer"}))
+    (let [max-tls (config/vice-max-initial-time-limit-seconds)]
+      (when (> tls max-tls)
+        (throw+ {:type  :clojure-commons.exception/bad-request
+                 :error (str "time_limit_seconds may not exceed " max-tls " seconds")}))))
```
psarando: There are already `validate-*` functions defined in this namespace. It might be more consistent to refactor this check into a `validate-time-limit` function?

psarando (approved):
LGTM 👍 
I just had a minor suggestion.

src/apps/service/apps/jobs/submissions.clj
ianmcorvidae: That's a good idea. The model was being uncooperative about not reimplementing things while I was working on this, not surprised it missed something, heh

PR #307, opened 2026-08-18
Update dependencies and gate on dependency conflicts

slr71:
Part of a fleet-wide pass over the DE Clojure repos to clear Leiningen's
"Possibly confusing dependencies" reports and bring dependencies current.
**20 warnings → 0.**

## Changes

- `build: normalize plugins and move editor tooling into profiles`
- `build: fail the build on confusing dependencies`

## The cider-nrepl leak is fixed at the source

Published `org.cyverse` poms used to declare `cider/cider-nrepl` at *compile*
scope, so a development REPL tool was landing on this service's runtime
classpath and inside its uberjar. Those libraries have since been re-released
with it removed, so bumping the pins was enough — **no exclusion workaround
needed**. Verified: 0 cider entries on the resolved classpath and 0 in the
built uberjar.

## The jackson pin is the important part of this PR

`apps` declares `metosin/compojure-api "1.1.14"` directly, and that drags
`cheshire 5.9.0 -> jackson-core 2.9.9`. By Maven's nearest-wins rule it beats
the much newer jackson the `org.cyverse` libraries bring transitively.

On `main` the family happens to resolve coherently at **2.17.2**. After the
dependency bumps it did not — it landed on `jackson-core 2.9.9` paired with
`databind 2.18.3`, an **eight-minor spread**. That surfaces as a runtime
`NoSuchMethodError`, not a resolution failure, and **`:pedantic?` cannot see it**
because each artifact is individually unambiguous. All 81 tests passed in that
state and the uberjar built fine, which is exactly why it is worth calling out.

The `:managed-dependencies` block pins the whole family back to `main`'s 2.17.2,
verified identical to `main` on the resolved classpath.

## Editor tooling moved out of top-level :plugins

`cider-nrepl` and `refactor-nrepl` were in shared top-level `:plugins`, putting
them on every invocation's classpath. They move to the `:repl` profile (and pick
up current versions). `clj-kondo` and `cljfmt` each get their own profile — both
have internally inconsistent trees that trip the gate on conflicts living
entirely inside third-party plugins.

Combined with the upstream releases, the uberjar no longer ships a development
REPL tool: **3 cider jars on the classpath -> 0, and 0 in the built jar**.

## Deliberately not changed

- `mvxcvi/clj-pgp` stays at **1.1.0** — the existing comment says 1.1.1 causes
  random decryption exceptions.
- `ring/ring-jetty-adapter` stays at **1.12.2**; 1.15.5 is a framework upgrade
  and belongs in its own PR.

## Extra verification

Largest test suite in the fleet, so the strongest signal available:

```
lein test     81 tests, 217 assertions, 0 failures
lein uberjar  builds; 0 cider entries
clj-kondo     clean
cljfmt        all source files formatted correctly
```

## Verification

```
confusing-dep warnings   20 -> 0
lein deps :tree          exit 0 (default, +dev,+test, +uberjar)
lein test                Ran 81 tests containing 217 assertions., 0 failures
lein uberjar             builds; 0 cider entries in the jar
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,27 +14,27 @@
             :url "https://cyverse.org/license"}
   :manifest {"Git-Ref" ~(git-ref)}
   :uberjar-name "apps-standalone.jar"
-  :dependencies [[org.clojure/clojure "1.12.4"]
+  :dependencies [[org.clojure/clojure "1.12.5"]
                  [clj-http "3.13.1"]
                  [com.cemerick/url "0.1.1" :exclusions [com.cemerick/clojurescript.test]]
                  [com.google.guava/guava "23.0"]
-                 [com.github.seancorfield/honeysql "2.7.1368"]
+                 [com.github.seancorfield/honeysql "2.7.1437"]
                  [dev.weavejester/medley "1.10.0"]
                  [metosin/compojure-api "1.1.14"]
-                 [org.cyverse/async-tasks-client "0.0.5"]
-                 [org.cyverse/authy "3.0.1"]
-                 [org.cyverse/clj-kondo-exports "0.1.0"]
-                 [org.cyverse/clojure-commons "3.0.12"]
-                 [org.cyverse/debug-utils "2.9.0"]
-                 [org.cyverse/kameleon "3.0.10"]
-                 [org.cyverse/mescal "4.1.0"]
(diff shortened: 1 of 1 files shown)


slr71:
I'm going to merge this now so that we can get these updates in as soon as possible. Reviews are still welcome

